PR #10, opened 2017-05-04
Configs

cmungall:
See https://github.com/biolink/biolink-api/issues/48


cmungall:
Note: still to actually wire up config objects

PR #15, opened 2017-05-16
some comment TODOs and a few small edits

dougli1sqrd:


cmungall:
ah for some reason I always confuse the fact that python doesn't support `x++` with a mistaken lack of support for `+=`

PR #26, opened 2017-05-17
WIP: Marshmallow defs

cmungall:
cc @dougli1sqrd


ontobio/assoc_schema.py
```diff
@@ -0,0 +1,48 @@
+from marshmallow import Schema, fields, pprint
+
```
dougli1sqrd: So then do you use these to produce classes? Are these like "meta" classes? Or do you instantiate these directly?

cmungall:
marshmallow.readthedocs.io/en/latest/

To instantiate a class:
https://marshmallow.readthedocs.io/en/latest/quickstart.html#deserializing-objects-loading

codegen for the class: ??

@kshefchek of relevance to phenopackets

cmungall:
Somewhat indirect: https://github.com/cwacek/python-jsonschema-objects

dougli1sqrd:
And then you'd have a User (from their example) class written somewhere that knows how to be instantiated with a **dict object, and then the @post_load decorator will actually instantiate the given class (ie User)?

PR #28, opened 2017-05-22
WIP: First pass at rdf generation

cmungall:


cmungall:
This PR covers RDF generation from the Association objects generated by the association parsers in gafparser.py. It's designed for immediate GO uses, to replace the existing OWLTools/Minerva translation, and there are 2 flavors generated:

 - GO-CAM aka LEGO degenerate models. See https://github.com/geneontology/minerva/blob/master/specs/owl-model.md
 - Simple OBAN-like association models

The evidence portion is shared.

Longer term goal is to share both modeling and code with dipper, cc @TomConlin @kshefchek @DoctorBud 

Note @balhoff has the inverse operations in java, lossy transformation from (complete) GO-CAM models to simple assoc formats, but this is not a simple inverting of operations due to the need to use inference. Ideally this could all converge more in the future.

ontobio/ontol.py
```diff
@@ -217,14 +217,43 @@ def filter_redundant(self, ids):
...
+        if 'subsets' in meta:
+            subsets = meta['subsets']
+        else:
+            subsets = []
+        if contract:
+            subsets = [self._contract_subset(s) for s in subsets]
+        return subsets
+
+    def _contract_subset(self, s):
+        if s.find("#") > -1:
```
dougli1sqrd: is there a reason to do this rather then `if "#" in s`?

tests/test_local_json.py
```diff
@@ -135,6 +135,19 @@ def test_graph():
     assert s1.pred == 'hasRelatedSynonym'
     assert s1.xrefs == ['GOC:mah']
 
+    GOSLIM = 'goslim_generic'
+    subsets = ont.subsets(NUCLEUS)
+    print("SUBSETS: {}".format(subsets))
+    assert GOSLIM in subsets
```
dougli1sqrd: are we not using the unittest framework? This would be much easier to test with unittest and nose (which easily runs tests).

kshefchek: +1 for a unit test framework, we use plain old python unittest in the dipper code:
https://docs.python.org/3/library/unittest.html

The future of nose is unclear, from their docs:
Nose has been in maintenance mode for the past several years and will likely cease without a new person/team to take over maintainership. New projects should consider using Nose2, py.test, or just plain unittest/unittest2.

ontobio/ontol.py
cmungall: no reason you should go ahead and fix these

tests/test_local_json.py
cmungall: Make a separate ticket for this. We should align across all our code. 

FWIW I've been happy with plain pytest; but it can be combined with unittest https://docs.pytest.org/en/latest/unittest.html

PR #42, opened 2017-06-22
Make parser defensive

cmungall:


cmungall:
Fixes https://github.com/geneontology/go-site/issues/376

PR #49, opened 2017-06-26
SemSim

cmungall:


ontobio/ontol.py
```diff
@@ -598,12 +598,49 @@ def logical_definitions(self, nid):
...
+            Node identifier for entity to be queried
+
+        Return
+        ------
+        dict
+            definition object, dict(val=TEXT, xrefs=LIST)
+        """
+        return self._get_meta_prop(nid, 'definition')
+        
+    def definition_val(self, nid):
```
dougli1sqrd: This seems like this should just be a method on whatever is returned from `self.definition`. So if you want the string version, you would just say `str(definition(id))` instead of providing a separate method here.

cmungall: the most common operation is getting the value; but in any case this will be dropped as I accidentlly reimplemented this code on a separate PR that's already been merged, doh

dougli1sqrd: Oh it's merged already? Where at? I was thinking we would release after this PR to get my logging changes in and this. 

Oh, I just saw that it's just a dictionary that's being returned and not a real Object. I guess my suggestion doesn't make as much sense then.

cmungall:
NOTE:

it's likely this will end up living in a different repo

PR #66, opened 2017-07-07
Filtered evidence file

dougli1sqrd:


dougli1sqrd:
Still testing, but seems to run on goa_chicken_complex-src.gaf with `--filter-out IEA` correctly, moving the non IEA lines to a different specified file, also while maintaining the full parsed version.

PR #72, opened 2017-07-11
Fixing some logic in report_ontology_id. …

dougli1sqrd:
This fixes the broken report tests.

### DIFF
--- a/ontobio/io/assocparser.py
+++ b/ontobio/io/assocparser.py
@@ -260,8 +260,13 @@ class AssocParser(object):
                                 msg="empty line")
             return ParseResult(line, [], True)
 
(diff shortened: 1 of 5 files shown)


dougli1sqrd:
Just for public's sake (I know we talked about this on skype). Having is_empty() is more expressive than just checking if the `ontology` is `None`. If in the `AssocParserConfig` we default create an empty `ontol.Ontology` object, we can then guarantee it's not None, and then have a more clear `is_empty()` to replace the `None` check. I think it's generally better to have "zero" objects that reflect that object being in a default or empty state: like an empty list `[]`, empty string `""`, etc since then we communicate to the reader that it's supposed to be a `list` or `str` etc.

PR #80, opened 2017-07-21
increasing buffer size for output file, only keeping some messages, n…

dougli1sqrd:
…o longer keeping any term ids

Check this out. I'm still waiting for uniprot to download to test locally, but you can have at it for now


kltm:
I'd defer to @cmungall .

kltm:
Looks fine to me though. Eager to test next release.

PR #102, opened 2017-09-12
uses assocwriter to write parsed validated gaf. parser replaces space s in relationship expressions with underscore

dougli1sqrd:
Ontobio just passes on spaces in the offending space in ids in relation expressions. This fixes this by simply replacing spaces with underscores. This won't create a correct id, as the space in the id is ultimately from upsteam. But this allows ontobio parsed gafs to then be turned into ttl at all, and ingested by blazegraph. We can correct the underscores with subsequent blazegraph loads (with the corrected ids), or with sparql queries later.

### DIFF
--- a/bin/ontobio-parse-assocs.py
+++ b/bin/ontobio-parse-assocs.py
@@ -147,7 +147,7 @@ def main():
     func(ont, args.file, outfh, p, args)
     if filtered_evidence_file:
         filtered_evidence_file.close()
-        
+
     if outfh is not None:
         outfh.close()
     if args.messagefile is not None:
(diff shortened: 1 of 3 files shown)


dougli1sqrd:
Okay so I had to remove the replace spaces with underscores. But the way we validate and write out the parsed gaf by using generators into a gafwriter will stay. This allows us to report errors (like spaces) correctly, but then still have the valid gaf lines.

kltm:
@dougli1sqrd Just to clarify, the raw GAFs are read in, these bad lines are noticed and filtered, and the scrubbed GAFs (with bad lines like these) are produced, correct?
@cmungall Are we actually *wanting* to attempt to fix these, or should they just be dropped? Are we doing other repairs to upstream? If so, we should have a central clearing house or documentation for what we're wiling to fix.

dougli1sqrd:
Yeah the bad gaf lines are filtered when building association dictionary objects. But the way filtered gaf was written out was just by copying the line read by the gaf. This meant that certain errors were still passed along (like spaces in ids when inside the relation expressions). So I decided to use the gaf writer to actually write gaf lines from a parsed gaf association dictionary object. This is more reliable because this correctly skips bad gaf lines more robustly. 

And yeah, my string replace is no longer there with my second commit. This just changes how we write out filtered gaf files now.

cmungall:
apologies for my emacs-generated whitespace. Should we have some kind of post-commit hook to lint the code? It makes it harder to see the actual semantic diff in the PR. I updated the title to reflect the diff.

The change makes sense, but this is actually quite a big change. We need to thoroughly check this before use in production. The writers have not been tested as thoroughly as the parsers. And it's not clear this fix preserves the headers. Have we done a diff of the output (for example on the mgi file) before and after?

PR #119, opened 2017-11-08
throw out evidence lines that have improper codes

dougli1sqrd:
Also this fixes a problem with the date, so that if it can be read by python, it is, but the written back out as the correct date format. Also, this adds the :graphType :gafCam triple to the ontology level triple for a gaf input file, along with a label of the input file. This lets these files be queried later on easier. Before this gets merged in we should decide once and for all the what I have called "metago" triple forms, as this will be in code.


dougli1sqrd:
Tests pass now.

PR #125, opened 2017-11-16
enhancement to filter by the produced_by information issue #545

selewis:
Includes test as well. Motivated by need to only obtain orthologs called by PANTHER


cmungall:
@deepakunni3 can this be closed?

deepakunni3:
@cmungall I am not sure about this PR. But the approach taken to filter is not ideal.
I would say lets close this, and create a new issue for the use case - filtering by `provided_by` field.

PR #146, opened 2018-02-26
simple bug fix

dougli1sqrd:


dougli1sqrd:
This skips header dictionary objects coming in from the reader, for the purposes of writing rdf from gaf.

PR #150, opened 2018-03-06
Enforce F/P/C aspect ONLY in GO

### LINKED ISSUE #138: GAF parsing should discard line and report if aspect is not F/P/C
See https://github.com/geneontology/go-annotation/issues/1797

cmungall:
Fixes #138


cmungall:
cc @putmantime

dougli1sqrd:
This is fine, it just seems a little overly specific to be in code. I think ideally this would be all controlled from a config file or something. But to get the others going, we should just merge this.

PR #152, opened 2018-03-07
add autocomplete and search for monarch

kshefchek:
- [x] Add lay person autocomplete
- [x] Add monarch autocomplete
- [x] Add monarch search
-  [x] fix lint warnings


kshefchek:
I've implemented autocomplete and slightly adjusted the previous search API call:
- adds prefix filter, boost function
- returns count of docs
- longest highlight

cmungall:
Looks good so far

Do we have a document that describes the various dependencies. What solr fields are expected, what are the expectations on how they are indexed, and how they get populated in the monarch pipeline

kshefchek:
I've used MagicMock to mock the PySolr.search() method so that, GolrSearchQuery.search(), GolrSearchQuery.autocomplete(), and GolrLayPersonSearch.autocomplete() can be tested.

It looks like we needed to specify the marshmallow version in setup.py, will rebase some of these commits to clean it up.

ontobio/sparql/sparql_ontology.py
```diff
@@ -122,9 +122,9 @@ def resolve_names(self, names, is_remote=False, synonyms=False, **args):
             results = set()
             for name in names:
                 results.update( self._search(name, 'rdfs:label', **args) )
-            if synonyms:
-                for pred in OIO_SYNS.values():
-                    results.update( self._search(name, pred, **args) )
+                if synonyms:
```
DoctorBud: It appears that there may be TAB characters here instead of SPACEs. Please check.

DoctorBud:
+1
looks good to me

Issue #154, opened 2018-03-10
assocparser QC check: check annotation not in blacklist subsets (gorule-0000008)


cmungall:
https://github.com/geneontology/go-site/issues/551

Different ontologies have different blacklist subsets; GO has 2

Documented here:
https://github.com/geneontology/go-site/blob/master/metadata/rules/gorule-0000008.md

Let's start with a GO-specific check at parse time. It won't affect other ontologies as only GO uses these subsets. 

We'd like to do this in SPARTA but until mechanism in place let's do this parse-time

dougli1sqrd:
Is this documented for GO terms by those terms in the SQL query? Or should we still use the property (gocheck_do_not_annotate) on the GO terms?

cmungall:
ignore the SQL: https://github.com/geneontology/go-site/pull/560

PR #156, opened 2018-03-13
gorule infrastructure and gorule 08 implementation

dougli1sqrd:
For https://github.com/biolink/ontobio/issues/154 and the future.

This provides an easy way to implement go rules that act on an annotation by annotation level to be checked by the gaf parser. This model can be extended if needed to other kinds of rule checks. The triggering of the QC in the gaf parser is small and easy to use. 

To add a Go Rule, just make a class like the GoRule08 existing example. This slightly mirrors the rule json format in go-site.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ foo:
 
 # only run local tests
 travis_test:
-	pytest tests/test_*local*.py tests/test_*parser*.py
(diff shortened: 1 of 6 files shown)


dougli1sqrd:
Yeah that makes sense. I think a solution for both of these points will become more clear as we put more weight on the rules. We can iterate as we come up with problems. 

I also see a future where this meshes closer to sparta also, but for now I think this is a good step.

PR #159, opened 2018-03-21
fixing config pathing and install issues

dougli1sqrd:


deepakunni3:
@cmungall I tried out this branch with biolink-api and there were no issues that I could find

dougli1sqrd:
Hmm, looks like a travis failure :(

PR #162, opened 2018-03-27
Fix condition for checking select_fields before adding GOlr specific fields

deepakunni3:
@nathandunn FYI

Here is the Gist of the JSON response: https://gist.github.com/deepakunni3/dd0698b4d2424c776aa89c5840a025c7


nathandunn:
Reviewed GIST and looks good.  Its all there!

deepakunni3:
@nathandunn Great! Thanks for checking

PR #163, opened 2018-03-29
Include result hits as 'numFound' in payload returned by GolrSearchQuery and GolrAssociationQuery

deepakunni3:
Add `numFound` into the response payload such that its available for association queries made by the biolink-api.

This PR aims to fix https://github.com/biolink/biolink-api/issues/155

@cmungall @putmantime


kshefchek:
+1

PR #167, opened 2018-04-19
Use correct go_context jsonld file in gaf->go-cam

cmungall:
NOTE: this is a temporary measure. We will build the go json ld context as part of the pipeline in future
See https://github.com/geneontology/go-site/issues/617


cmungall:
@dougli1sqrd, after you merge this and snapshot runs, hand over the jrnl to @lpalbou and @dustine32 for testing

cmungall:
@dougli1sqrd will make a new release so this goes into snapshot

PR #170, opened 2018-04-26
Invert subject and object from Solr results when invert_subject_object=true

deepakunni3:
When `invert_subject_object=True`, the Solr query is made after inverting the subject and the object, but the results from Solr are not parsed properly.

This PR aims at ensuring that the subject and object ordering is maintained in the response parsed from Solr results.

Required for biolink/biolink-api#143

### DIFF
--- a/ontobio/golr/golr_associations.py
+++ b/ontobio/golr/golr_associations.py
@@ -293,10 +293,10 @@ def calculate_information_content(**kwargs):
 
 from ontobio.vocabulary.relations import HomologyTypes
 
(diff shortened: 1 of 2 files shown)


deepakunni3:
@kshefchek Could you take a look at this PR when you have time.
The changes mostly occur in `golr_query.py`. It would be great to get your feedback on the changes made and the approach used.

deepakunni3:
@cmungall This PR is ready to merge but I had some merge conflicts, which I had to resolve with `git merge upstream/master`.
Let me know if there is a better way to do this.

I'll let you merge it once its verified.

PR #171, opened 2018-04-29
Paintmerge 591

dougli1sqrd:
The main change here is to add a validate.py script in the bin directory. This orchestrates much of what the old makefile did. You give validate the target directory, the metadata directory, and which group to produce data from:

First it finds the <group>.yaml. For each gaf dataset in the group yaml, download the source gaf, and run it through the gaf validation step, producing the noiea, and standard reports we all know and love. We call this <dataset>_valid.gaf.

Then we produce a gpi file and a gpad file from the validated gaf.

Next we check if there is a corresponding paint_<group> dataset. If so, we download the src paint gaf, validate it, but use the earlier GPI to correct entitites in the paint gaf. 

Once we have a valdiated paint gaf, we do the "merge", where we take validated paint gaf and just paste it on the bottom of the standard validated gaf. This is then the final <dataset>.gaf product. 

Then we produce a ttl on this final gaf product. 

Along the way, each file is zipped.

There are few outside changes to ontobio, but they're few. Travis tests pass. I have a test branch in go-site where I have a new makefile that uses this ontobio. Before we can do a real jenkins test on the iteration branch, we will probably have to merge and release this, yes?

### DIFF
--- a/bin/rdfgen.py
+++ b/bin/rdfgen.py
@@ -13,7 +13,7 @@ def cli():
 @click.option("--association", "-a", default="gaf", type=click.Choice(["gaf"]))
 @click.option("--ontology", "-r", type=click.Path(exists=True, readable=True, dir_okay=False))
 @click.option("--output", "-o", type=click.File('wb'))
-@click.argument("association_file", type=click.File('r'))
+@click.argument("association_file", type=click.Path(exists=True))
 def convert(association, ontology, output, association_file):
     click.echo("converting {}".format(association))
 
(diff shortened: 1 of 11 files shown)


kltm:
@dougli1sqrd can you do a pre-release and do a small pipeline run with this?

dougli1sqrd:
@kltm Yeah, I'd love to. What's your suggestion here though? In order for go-site/pipeline to see ontobio at all, wouldn't we have to do a pypi release? The best thing we can do is if pipeline can install ontobio at this branch directly from git. Would that be hard?

kltm:
If you can release and point to it in a requirements.txt, why not just make an RC release?

PR #173, opened 2018-05-02
Support a list of prefix filters (pos and neg)

kshefchek:
For autocomplete and search, we want the ability to filter given a list of prefixes.  This use case is required for the gamma reasoner.

### DIFF
--- a/ontobio/golr/golr_query.py
+++ b/ontobio/golr/golr_query.py
@@ -415,10 +415,18 @@ class GolrSearchQuery(GolrAbstractQuery):
             params['fq'] = []
 
         if self.prefix is not None:
-            if self.prefix.startswith('-'):
-                params['fq'].append('-prefix:"{}"'.format(self.prefix[1:]))
-            else:
-                params['fq'].append('prefix:"{}"'.format(self.prefix))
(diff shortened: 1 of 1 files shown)


DoctorBud:
+1
Looks fine to me.
If we have unit tests for BioLink, they should probably be extended to verify this new capability (don't stop the PR tho).

kshefchek:
@DoctorBud good point on testing.  This particular bit of code is untested right now, but curious your thoughts on how to go about this.

Currently we have some unit tests for this class here:
https://github.com/biolink/ontobio/blob/e6e0fc480b696249b57816f2a9210590fc929889/tests/unit/test_golr_search_query.py

I'm testing the solr response by mocking the pysolr.search function to return a Results object I've created as a test fixture (as json, which maybe isn't the best choice).

We also have some integration tests that test the live monarch and go solr instances.  We could do this here as well.

Another alternative is using docker compose to create a test environment with solr and ontobio, load some test data and test ontobio.  I've done something similar here: https://github.com/monarch-initiative/hpo-plain-index  It's a little more work up front but then we won't get failures if a service is down.

PR #174, opened 2018-05-08
increase version of prefixcommons

cmungall:
This ensures that by default shortest URIs are chosen if multiple
contractions are possible


kshefchek:
+1

PR #184, opened 2018-05-31
slices off extra columns and proceeds with a warning if we see more t…

dougli1sqrd:
…han 17


kltm:
@cmungall Severe, but what could go wrong?

cmungall:
this is fine as first pass

we should check if we are truncating non-blank columns. If so, it is more severe and should be reported.

kltm:
Note relation to https://github.com/geneontology/go-annotation/issues/1997#issuecomment-393012816

PR #189, opened 2018-06-15
removed gorule8 as an ontobio level check as it should be soft, and exists as a sparql rule

dougli1sqrd:


dougli1sqrd:
https://github.com/geneontology/go-site/issues/679

PR #196, opened 2018-07-10
added taxon triple to gp classes

dougli1sqrd:


dougli1sqrd:
@cmungall How does this look for you now?

PR #202, opened 2018-07-19
No ibas in gpad

dougli1sqrd:
This replaces evidence codes with the CURIE ECO id.

This also correctly reads the combined GAF product as PAINT, so that the merged IBA paint annotations do not get filtered. See https://github.com/geneontology/pipeline/issues/44 for more info.


kltm:
Okay. Let's try and check this on experimental first.

Issue #205, opened 2018-07-26
GAF Columns that process pipe separated IDs should drop entire line on any one bad ID


dougli1sqrd:
The GAF parser parses certain columns that are pipe separated IDs a little differently than other columns. For example, the reference column processes pipe separated PMID IDs: `PMID:12345|PMID:54321`. 

Currently, if any ID is found to be invalid (say because of bad characters in the ID) it is dropped, but the line remains. 

The bug is when there is one ID with an error, the ID is dropped, essentially making columns processed in this way potentially disappear. For example, if the reference column has:
`PMID:1234[2]` this gets converted by the parser into an empty list. 

We should drop the whole line if we see an error in any one pipe separated ID.

dougli1sqrd:
https://github.com/biolink/ontobio/pull/207

PR #207, opened 2018-07-27
bad ids in pipe fields now filter out the line. Tests for the new pipe logic

dougli1sqrd:
https://github.com/biolink/ontobio/issues/205

### DIFF
--- a/ontobio/io/assocparser.py
+++ b/ontobio/io/assocparser.py
@@ -513,11 +513,11 @@ class AssocParser(object):
 
         return True
 
-    def validate_pipe_separated_ids(self, column, line, empty_allowed=False) -> Optional[List[str]]:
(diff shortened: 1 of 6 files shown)


cmungall:
withfrom - both pipes and commas are allowed (disjunction and conjunction)

As a first pass in this PR we should use commas as an alternate delimiter. The validate-pipe method can take an additional argument for additional delimiters.

Note this will not be fully information preserving as we will flatten disjunctions and conjunctions one list. this is fine for this PR

For full information presentation we can include an additional field in the association object that is either a structured object or the original string.

dougli1sqrd:
The GAF Writer will then join each with/from element with a pipe when written back out, is this okay?

foo,bar|hello,world ==> foo|bar|hello|world

cmungall:
Yes, this is fine for this PR

Issue #219, opened 2018-08-15
Change JSON file reports during GAF parsing to support better reporting


kltm:
The current JSON output is lacking lacking in a few key aspects to support the reporting that we'd like to have in the GO.

We'd like to have the following changes to the output:

- [x] the top-level fields include `group` (resource)
- [x] the violations are grouped by either their rule (id) or a catch-all
- [ ] the violations include `species` (taxon id)

(I may be forgetting one.)

There may be an arbitrary cut-off per rule of listed violations.

Tagging @dougli1sqrd and @cmungall

dougli1sqrd:
```
{
    "group": "goa_chicken_complex",
    "lines": 1234,
    "skipped_lnes": 24,
    "messages": {
        "other": [
            {
                "level": "WARNING",
                "line": "Hello\tworld\tgoterm\tetc",
                "type": "INVALID_ASPECT",
                "message": "Wrong aspect",
                "obj": "GO:1234567",
                "taxon": "NCBITaxon:12345"
            },
            {
                "etc": "etc"
            }
        ],
        "gorule-0000001": [
            {
                "level": "ERROR",
                "line": "not\tenough\tcolumns",
                "type": "Wrong number of columns in this line",
                "message": "Line has 3 columns, when it should have 15 or 17",
                "obj": "",
                "taxon": ""
            }
        ]
    }
}
```
I do something like this.

dougli1sqrd:
So, if a gaf has more than one taxon, which one do I use? Do I use all of them? @cmungall @kltm

dougli1sqrd:
Hello @kltm and @cmungall,

This PR for ontobio does this: https://github.com/biolink/ontobio/pull/226
This PR for go-site cleans up any issues with the new format: https://github.com/geneontology/go-site/pull/779

We have a successful build in Jenkins: https://build.geneontology.org/job/geneontology/job/pipeline/job/json_format_test/4/ which uses this code.

This is the output combined report: http://skyhook.berkeleybop.org/json_format_test/reports/combined.report.json

And this is an example of WB report.json: http://skyhook.berkeleybop.org/json_format_test/reports/wb.report.json

Let's have discussion here. If this is good, let's merge and release ontobio, and then at the same time merge the go-site PR.

kltm:
I think the question for me is: if we take this now, will be be able to back out or fix quickly enough to deal with this month's release?

dougli1sqrd:
[...]

PR #226, opened 2018-08-23
new report json

dougli1sqrd:
Sorry there's so much here.

There were a lot of moving parts that had to shift around.


dougli1sqrd:
https://github.com/biolink/ontobio/issues/219

PR #231, opened 2018-08-29
Revert "gorule11"

cmungall:
Reverts biolink/ontobio#216


dougli1sqrd:
Won't merge. Will fix existing rule.

PR #233, opened 2018-08-31
making sure all gaf cam models have rdfs:labels

dougli1sqrd:
I hadn't set the `label` to the filename of the incoming gaf for some reason. With it set, you can see in https://github.com/biolink/ontobio/blob/master/ontobio/rdfgen/assoc_rdfgen.py#L62 that with the label definitely set, we add an rdfs:label triple. 

Tested with a previously broken MGI, we get the label.


kltm:
Looks good to me.

Issue #234, opened 2018-08-31
Raise logging level of Ontology.init message


dustine32:
This line in ontol.py:
```
logging.info(‘Graph initialized, nodes={}’.format(self.graph.nodes()))
```
..is flooding the AGR loader's logs when setting the log level to INFO. Suggesting that this message logging be raised to DEBUG level and also maybe use `logging.getLogger(__name__)` in ontol.py for better logging control.

I'll put in the PR shortly.

PR #235, opened 2018-08-31
Issue 234 - ontol logging stuff

dustine32:
For issue #234 . Let me know if the getLogger usage affects anything huge.


dustine32:
Not too nitpicky! I was probably too lazy to change all of the `logging` variables and/or worried that changing the `logging` naming would ruin convention. I'll update the file to use `logger` instead to remove any uncertainty.

dustine32:
Thanks @dougli1sqrd !

PR #239, opened 2018-09-12
boost kw fields for curies

kshefchek:
See https://github.com/monarch-initiative/monarch-app/issues/1625 as background.

When a user enters a string with a colon we assume it is an identifier and boost the id_kw and equivalent_curie_kw fields.  The downside is that if we want to autocomplete or search on labels with a colon these will be down weighted.

Also removing the conf dir to avoid confusion.


kshefchek:
We do have some [labels that contain colons](https://solr-dev.monarchinitiative.org/solr/search/select/?defType=edismax&qt=standard&indent=on&wt=json&start=0&fl=label_std,score&facet=true&facet.mincount=1&facet.sort=count&json.nl=arrarr&facet.limit=25&hl=true&hl.simple.pre=%3Cem%20class=%22hilite%22%3E&hl.snippets=1&facet.field=id&q=*\:*&qf=label_std^1&hl.fragsize=0) but they don't seem to be likely searched on, the top returned items:

1. GO:molecular_function
2. Characterization of the Han:SPRD rat model for hereditary  polycystic kidney disease
3. A canine model of hemophilic (factor VIII:C deficiency) bleeding.
4. Water and feed intake, feed efficiency, and abdominal fat levels of dwarf and normal chickens selected under different water:feed ratio environments
5. EEG characterization of audiogenic seizures in the hamster strain GASH:Sal.
6. A Polymorphism in the Melanocortin 4 Receptor Gene (MC4R:c.92C>T) Is Associated with Diabetes Mellitus in Overweight Domestic Shorthaired Cats.
7. Presence of an Acyl Carrier Protein in NADH:ubiquinone  Oxidoreductase from Bovine Heart Mitochondria

HGVS variant symbols may potentially be an issue, but in testing "MC4R:c.92C>T" even with the curie boost this shows up first.
"BRCA1:c.4332T>C (p.Asn1444=)" also returns the correct result.  So the additional curie boost doesn't seem like an issue for variants.

"NADH:ubiquinone" and "GASH:Sal" also still return the correct results with the curie weight.

kltm:
@kshefchek At least for number one, that seems to be an issue with an imported term clobbering the proper GO one. @cmungall for details.

PR #243, opened 2018-09-13
Required for new GO queries in BioLink

lpalbou:
Some GOlr queries will probably be later replaced to more correctly use GolrSearchQuery.
Some additional SPARQL queries.

PR required for following PR in BioLink

### DIFF
--- a/ontobio/golr/golr_query.py
+++ b/ontobio/golr/golr_query.py
@@ -1610,3 +1610,76 @@ class GolrAssociationQuery(GolrAbstractQuery):
             # default to input
             return id
         return ids[0]
+
+
+
+
+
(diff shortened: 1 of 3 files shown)


ontobio/golr/golr_query.py
```diff
@@ -1610,3 +1610,76 @@ def map_id(self,id, prefix, closure_list):
             # default to input
             return id
         return ids[0]
+
+
+
+
+
+### This may quite possibly be a temporary code, but it looks a lot simpler than the above for more customizable Solr queries
+import requests
```
deepakunni3: As per discussion, would be good to move this out of golr_query and into utils

PR #252, opened 2018-10-02
adding back gorule 8, with warnings

dougli1sqrd:


kltm:
Let's punt on this until we can get a release out the door, unless this is a critical fix.

dougli1sqrd:
Sure thing. We would have to release and then update for this to get included in a new pipeline run anyway.

kltm:
Seems fine when the conflict is deescalated.

PR #255, opened 2018-10-06
add owlsim2 wrapper to drive monarch-app

kshefchek:
The goal of this PR is to replicate and replace the semantic search and phenotypic sufficiency scoring in the legacy monarch-app javascript API, specifically:

[compareEntities](https://github.com/monarch-initiative/monarch-legacy/blob/70c790/lib/monarch/api.js#L1218) -  Given a query and one or more target entities (such as a disease, genotype, gene), this will map them to their phenotypes and perform a phenotypic comparison

[searchByPhenotypeProfile](https://github.com/monarch-initiative/monarch-legacy/blob/70c790/lib/monarch/api.js#L948) - Returns phenotypically similar entities of a given type

[getInformationProfile](https://github.com/monarch-initiative/monarch-legacy/blob/70c790/lib/monarch/api.js#L816) - Compute annotation sufficiency score


kshefchek:
Combing through the owltools code and js implementation some notes:
1. Weighted negation is commented out for the simple score: https://github.com/monarch-initiative/monarch-legacy/blob/70c790/lib/monarch/api.js#L899, but the implementation makes sense.  I'm unsure why the code commented but I will include negation support in the ontobio code.
2. Negation is included in the categorical score, but is not weighted.  I would think we should down weight negated phenotypes, similar to above. See https://github.com/monarch-initiative/monarch-legacy/blob/70c790/lib/monarch/api.js#L906

cmungall (approved):
This is all clear and makes sense, thanks, merge when ready

PR #256, opened 2018-10-08
DOIs are no longer filtered

dougli1sqrd:


dougli1sqrd:
https://github.com/geneontology/go-site/issues/836

ontobio/io/assocparser.py
```diff
@@ -492,7 +492,8 @@ def _validate_id(self, id, line: SplitLine, allowed_ids=None, context=None):
             self.report.error(line.line, Report.INVALID_ID, id, "GORULE:0000027: must be CURIE/prefixed ID", rule=27)
             return False
 
-        if AssocParser.non_id_regex.search(id.split(":")[1]):
+        # we won't check IDs with doi prefix, everything else we want to check
+        if not id.lower().startswith("doi") and AssocParser.non_id_regex.search(id.split(":")[1]):
```
kltm: I'd argue that this is a fairly wasteful set of state machines. Doing something like a pre-compiled `/^doi\:/i` would get both without conversions and multiple calls.
On the side, I'm not sure DOI (upper) is a "correct" CURIE. It would be interesting to know if both are used.

kltm: I mention this as we have been talking about things that can boost the time marginally, but have large aggregate effects.

dougli1sqrd: Oh that makes sense. As far as correct upper/lower of DOI IDs, I think capitalized is definitely used (as seen here http://snapshot.geneontology.org/reports/fb-report.html#gorule-0000027). But perhaps we just should call it, and say that only lower case `doi` is allowed? Not sure what the right answer is there. (@cmungall, tagging in case you have a chance to weigh in at some point)

kltm: Go ahead and be insensitive for now, we can tune it later--we at least know what the intention was and can either fix or just deal.

dougli1sqrd:
@kltm I'm going to merge this unless there's anything else, thanks for the feedback earlier.

kltm:
NP. We'll revisit everything later.

PR #260, opened 2018-10-12
changing the ordering of the MGI double prefix patch to fix other dou…

dougli1sqrd:
…ble prefix IDs


kltm:
Assuming this code is db-xrefs lookup only.

PR #262, opened 2018-10-25
fixing subject and object extensions

dougli1sqrd:


kltm:
@dougli1sqrd This includes tests for passing through the isoform as-is?

PR #264, opened 2018-10-29
interacting taxon now parsed and passed through

dougli1sqrd:


kltm:
Looks good. Is there a test got `interacting_taxon` ingest?

dougli1sqrd:
The round trip test in `test_assoc_writer` has an interacting taxon

kltm:
Let's give it a whack.

PR #265, opened 2018-10-30
adding test_assoc_writer to makefile, correcting bugs re isoform output

dougli1sqrd:


dougli1sqrd:
This is @kltm: looks good; we'll retry tonight.

PR #270, opened 2018-11-07
fixing the taxon issue

dougli1sqrd:


dougli1sqrd:
From the gitter chat:
So, I’m still drilling into what’s going on with the pipeline missing date issues. For reference, normal process is we go from src —> valid —> gpi, where we use the gafparser to “re parse” the validated src. So: src —> [gafparser(src)] —> valid —> [gafparser(valid)] —> gpi. Something is happening in the second gaf parser on the valid data. In WB for example it ends up filtering all but 124 lines of the valid. This is what I’m drilling into see what the deal is.


okay, I think I know what the issue is. It’s the recent interacting taxon change. The code that writes the taxon out is including the pipe somehow? I’m investigating further what specifically the issue is. So, single taxons are being written out like taxon:1234| which fails the gaf parse step the second time around

PR #277, opened 2019-01-16
Networkx 2.0 migration

deepakunni3:
There are a few changes between networkx 1.x and networkx 2.x

This PR upgrades networkx dependency from 1.11 to 2.2.
It also makes the necessary changes across the Ontobio codebase.

**The official migration documentation:** https://networkx.github.io/documentation/stable/release/migration_guide_from_1.x_to_2.0.html

**Lessons learned (an unofficial migration documentation):**

* `graph.nodes_iter()` is now replaced with `graph.nodes()`
* `graph.edges_iter()` is now replaced with `graph.edges()`
* `graph.nodes()` will return an NodeView. Convert this to a list via `list(graph.nodes())`
* `graph.edges()` will return an EdgeDataView. Convert this to a list via `list(graph.edges())`
* `graph.edges()` will return a tuple of edges, `(u,v)`
* `graph.edges(data=True)` will return a 3-tuple, `(u,v,d)`
* `graph.add_node()` does not have `attr_dict` argument anymore. Instead, supply all properties of a node as keyword arguments. Similarly for `graph.add_edge()`
* To add a new property to a node,
```py
graph.nodes[n]['property_name'] = value
```

* To add a new property to an edge,
```py
# for a DiGraph
graph[u][v]['property_name'] = value
# for a MultiDiGraph
graph[u][v][0]['property_name'] = value
```

* Accessing edge properties,
```py
graph.get_edge_data(u, v)
```

* `networkx.neighbors()` method has been moved to `graph.neighbors()`

### DIFF
--- a/ontobio/io/ontol_renderers.py
+++ b/ontobio/io/ontol_renderers.py
@@ -219,7 +219,7 @@ class SimpleListGraphRenderer(GraphRenderer):
         for n in ontol.nodes():
             s += self.render_noderef(ontol, n, **args) + "\n"
             #for n2 in ontol.parents(n):
-            #    for _,ea in g[n2][n].items():
+            #    for _,ea in g.get_edge_data(n2,n).items():
             #        s += '  {} {}'.format(str(ea['pred']), self.render_noderef(ontol, n2, **args))
             #        s += "\n"
         return s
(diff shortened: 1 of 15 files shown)


cmungall:
A heroic effort!

We should bump up the minor version of ontobio for this

deepakunni3:
@cmungall Thanks!

Yes, that is a good idea. New version would be 1.7.0

dougli1sqrd:
I can release a new ontobio 1.7.0 when this hits master

deepakunni3:
@dougli1sqrd Thanks! 👍

PR #278, opened 2019-01-17
Add max ic to sim result object

kshefchek:
Global max IC is needed for phenogrid, adding metadata field with max_max_ic and updated the tests as needed.  Also added cachier to owlsim calls.

### DIFF
--- a/ontobio/model/similarity.py
+++ b/ontobio/model/similarity.py
@@ -42,6 +42,14 @@ class TypedNode(Node):
         self.type = type
         self.taxon = taxon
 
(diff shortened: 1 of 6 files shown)


DoctorBud:
+1

kshefchek:
In order to get caching to work a few methods have to be moved outside of the owlsim2 API class.  Caching doesn't appear to work for instance methods, and I'm not able to apply the staticmethod decorator and cachier decorator.  @deepakunni3 @cmungall curious if you have ran into this?

Issue #203, opened 2018-07-23
Add GO aspect convenience methods


dustine32:
I keep finding myself rewriting the same is_mf, is_cc, is_bp, or get_aspect(term) functions. It'd be cool if ontobio just had these methods ready to go. I can add these myself.

Basic idea, which I could be oversimplifying so let me know, is to check a term's ancestors for presence of the root term. Example:

```
def is_molecular_function(go_term):
    mf_root = "GO:0003674"
    if go_term == mf_root:
        return True
    ancestors = ontology.ancestors(go_term)
    if mf_root in ancestors:
        return True
    else:
        return False
```

PR #281, opened 2019-01-23
Added GO aspect convenience methods for #203

dustine32:
Let me know if these are in the wrong place (`ontol.py`) or they're just, plain wrong.


deepakunni3:
@cmungall Would `get_isa_partof_closure` be a better name for that method? It seems to do exactly that.
We could also add `get_isa_closure` along with `get_isa_partof_closure`

ontobio/ontol.py
```diff
@@ -358,6 +358,73 @@ def node_type(self, id):
...
 
+    def get_ancestors(self, go_term):
+        """
+        Returns the ancestors from the is_a, part_of relation traversed GO subontology of go_term's ancestors.
+
+        subontology() primarily used here for speed when specifying relations to traverse.
+        """
+        ### BFO:0000050 = part of
+        ### I assume "subClassOf" = is a?
+        all_ancestors = self.ancestors(go_term)
```
deepakunni3: @dustine32 

If you call `self.ancestors(go_term, reflexive=True)` then you won't have to do `all_ancestors.append(go_term)` as the term itself is included in the list of ancestors returned by `self.ancestors()`.

dustine32: Thanks @deepakunni3 ! I'll try that option out.

dustine32:
Cool, thanks for the feedback! I'll reorganize this into a file under `utils/` and make those changes.

I do remember running into the MF-part_of-BP issue for this a while ago, causing MF terms to be erroneously identified as BPs. But I got around it by just asking if it was an MF first, though `is_biological_process` would still be left saying an MF's a BP. I'll try limiting this to just use the 'is a' relation.

ontobio/ontol.py
```diff
@@ -358,6 +358,73 @@ def node_type(self, id):
         """
         return self.node(id)['type']
 
+    def get_ancestors(self, go_term):
+        """
+        Returns the ancestors from the is_a, part_of relation traversed GO subontology of go_term's ancestors.
+
+        subontology() primarily used here for speed when specifying relations to traverse.
```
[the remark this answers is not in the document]
dustine32: Yeah, I figured this out a long time ago so it might've changed (or cache is messing me up), but specifying relations without using `subontology` first will calculate the filtered graph using the whole GO ontology rather than just the smaller subset of "all relations" ancestors of a node.

Like, specifiying relations requires looking at edge properties (in `ontol.get_filtered_graph()`) whereas not just cares that an edge is there. Does this make sense?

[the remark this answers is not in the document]
deepakunni3: Thank you @dustine32 
Yes, I see the difference and where the bottleneck is 👍

ontobio/ontol.py
dustine32: Yep, it worked!

dustine32:
I think these changes are ready to go. I now have this in `util/go_utils.py` and I'm only traversing the graph through is_a relations. @cmungall @deepakunni3 Let me know if I missed anything. Thanks!

dustine32:
Awesome thanks!

PR #285, opened 2019-01-31
Upgrade pyyaml req due to vulnerability

dustine32:
This was causing the annoying security vulnerability warning to appear in some of my repos due to a pyyaml version < 4.2b1 (3.13) being installed through ontobio.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -46,7 +46,7 @@ setuptools.setup(
     install_requires=[
         'networkx==2.2',
         'jsobject',
-        'pyyaml==3.13',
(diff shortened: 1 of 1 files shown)


dustine32:
Oh fun! I see the error at the bottom:
```
error: PyYAML 4.2b4 is installed but pyYaml==3.13 is required by {'yamldown'}
```
@dougli1sqrd Are you able to update this req in `yamldown` or do you think we should just panic and give up?

dougli1sqrd:
ope, yep, I can fix that

PR #289, opened 2019-02-06
validate.py gets an option to pass a base URL for relative download paths

dougli1sqrd:
This will also throw an error if validate finds a relative path but isn't supplied a base path:

Path in gocam config:
`source: ./products/annotations/noctua_mgi.gpad.gz`

Note in the command the option `--base-download-url` is not supplied
```
$ python bin/validate.py produce gocam -m ~/lbl/geneontology/go-site/metadata/ -t target/ -o go-ontology.json
Making products gaf, gpi.
Products will go in /Users/edouglass/lbl/biolink/ontobio/target
Found gocam metadata at /Users/edouglass/lbl/geneontology/go-site/metadata/datasets/gocam.yaml
Loading ontology: go-ontology.json...
Found noctua_mgi
Downloading source gaf to /Users/edouglass/lbl/biolink/ontobio/target/groups/gocam/noctua_mgi-src.gaf.gz
Error: Option --base-download-url was not specified and the config url ./products/annotations/noctua_mgi.gpad.gz is a relative path.
```

### DIFF
--- a/bin/validate.py
+++ b/bin/validate.py
@@ -81,19 +81,27 @@ def rule_id(rule_path) -> str:
     return os.path.splitext(os.path.basename(rule_path))[0]
 
 
-def download_source_gafs(group_metadata, target_dir, exclusions=[]):
+def download_source_gafs(group_metadata, target_dir, exclusions=[], base_download_url=None):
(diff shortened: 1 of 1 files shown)


kltm:
I am now clear--looks good.

PR #290, opened 2019-02-06
basic skeleton of an internal Association object

dougli1sqrd:
This PR just codifies what we've been doing in the dictionary for a long time now. We could do some minimal refactoring over time to convert the parsers and writers to use this Association object rather than the dictionary. This ensures all parsers and writers are on the same page. 

With @dustine32 and I working on the code concurrently where gpad and gafs will need to ensure solid communication, we should have a typed go-between structure. This is just the same as the dictionary, just named now.

This particular PR doesn't change any behavior - Dustin and I will slowly add this in where relevant over time in the parser/writer. @dustine32 if there's anything else we need to add to the model to make your work okay, let's make sure we know about it here. And @cmungall, did I capture your intent with the original json/dict structure okay here? 

My biggest question is about the Subject Extensions. In the GAF spec at least, it says we can only have 0 or 1 entries of this. But in the dictionary this is a list. Since I also know you are going for a higher level generalized form here for the subject extensions, I wanted to get your input that this is per your envisioning.

### DIFF
--- a/ontobio/model/association.py
+++ b/ontobio/model/association.py
@@ -51,8 +51,8 @@ class ExtensionExpression:
     conjunctions: List[ExtensionConjunctions]
 
 @dataclass(repr=True, unsafe_hash=True)
-class Association:
-    source_line: str
+class GoAssociation:
+    source_line: Optional[str]
     subject: Subject
     relation: Curie # This is the relation Curie
     object: Term
@@ -66,6 +66,3 @@ class Association:
     provided_by: Provider
     date: Date
     properties: Dict[Curie, List[str]]
-
-    def __str__(self):
-        return self.source_line
--- a/ontobio/model/association.py
+++ b/ontobio/model/association.py
@@ -42,15 +42,14 @@ class ExtensionExpression(NamedTuple):
     """
     conjunctions: List[ExtensionConjunctions]
 
-@dataclass(repr=False, unsafe_hash=True)
-class Association:
+class Association(NamedTuple):
     source_line: str
     subject: Subject
     relation: Curie # This is the relation Curie
     object: Term
     negated: bool
     qualifiers: List[Curie]
-    aspect: Aspect
+    aspect: Optional[Aspect]
     interacting_taxon: Optional[Curie]
     evidence: Evidence
     subject_extensions: List[ExtensionUnit]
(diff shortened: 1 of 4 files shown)


cmungall:
Can we stay in python3.6 for now? I have used dataclasses with 3.6
successfully

>

dustine32:
I tried importing the `Association` object after switching back to python3.6.4_4 and got an error:
```
$ brew switch python3 3.6.4_4
$ python3
>>> from ontobio.model.association import Association
>>> /Users/ebertdu/go/ontobio/env/lib/python3.6/site-packages/cachier/mongo_core.py:24: UserWarning: Cachier warning: pymongo was not found. MongoDB cores will not work.
  "Cachier warning: pymongo was not found. MongoDB cores will not work.")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/ebertdu/go/ontobio/ontobio/model/association.py", line 45, in <module>
    @dataclass(repr=False, unsafe_hash=True)
NameError: name 'dataclass' is not defined
```
Do you have to add spices to make dataclasses work in 3.6 or am I importing this wrong?

dustine32:
Oh maybe we need to install this module https://github.com/ericvsmith/dataclasses ?

It's a backport of dataclasses for 3.6

cmungall:
Just import dataclasses

ontobio/model/association.py
```diff
@@ -0,0 +1,63 @@
+import json
+import typing
+import collections
+import enum
+import datetime
+
+from typing import List, Optional, NamedTuple, Dict
```
dustine32: Added this `Dict` to typing. Was getting an error at the `Association.properties` declaration.

dougli1sqrd:
Yeah if we want 3.6 python, let's use the https://pypi.org/project/dataclasses/ import in the setup.py

dustine32:
Last commit 796ea75 removed `dataclasses` import and `setup.py` dependency since we switch to `NamedTuple`.

cmungall:
We should be using dataclasses

dougli1sqrd:
@cmungall I updated to GoAssociation and made source_line optional

dougli1sqrd:
@cmungall I'd like to poke you about this PR. I think I made all the changes you suggested.

dougli1sqrd:
I chatted with @kltm and we'll merge this in.

PR #297, opened 2019-02-28
Specify PyYaml version

kshefchek:
We should fix our yaml.load calls -  https://github.com/yaml/pyyaml/wiki/PyYAML-yaml.load(input)-Deprecation

But in the short term this will allow to run biolink-api


DoctorBud:
+1

PR #300, opened 2019-03-07
allowing unzipped upstream sources

dougli1sqrd:
If config `compression` field is empty, then we assume no compression.

### DIFF
--- a/bin/validate.py
+++ b/bin/validate.py
@@ -80,15 +80,30 @@ def gorule_title(metadata, rule_id) -> str:
 def rule_id(rule_path) -> str:
     return os.path.splitext(os.path.basename(rule_path))[0]
 
+def source_path(dataset_metadata, target_dir, group):
+    extension = dataset_metadata["type"]
+    if dataset_metadata["compression"]:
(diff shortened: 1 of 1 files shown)


dougli1sqrd:
The test config for reactome to demonstrate the success:
```
id: reactome.gaf
   label: "reactome gaf file"
   description: "gaf file for reactome from Reactome - a curated knowledgebase of biological pathways"
   url: http://current.geneontology.org/annotations/reactome.gaf.gz
   type: gaf
   dataset: reactome
   submitter: reactome
   compression:
   source: https://reactome.org/download/current/gene_association.reactome
   entity_type:
   status: inactive
   species_code:
   taxa:
```
Note the blank compression.

kltm:
https://github.com/geneontology/go-site/commit/7cbd3e444c17#commitcomment-32646156
@dougli1sqrd For the final product, we'll need to make sure that the final *-src.gaf.gz byproducts are compressed as well, not just copied.

dougli1sqrd:
@kltm with my latest update uncompressed sources are now also zipped.

kltm:
Talked with @dougli1sqrd , feeling good about this one.

PR #304, opened 2019-03-19
command line option to suppress messages from rules with specified tags

dougli1sqrd:
`validate.py produce` now has an option `--suppress-rule-tag tag`. This will suppress messages generated from rules with the tag `tag`. Default is no suppression of messages. The canonical thing to do will be to use this with `silent`, which gorule 26 currently has - which will suppress markdown report of rule 26. 

See also https://github.com/geneontology/go-site/issues/884


kltm:
Looks as described by our talk last week. Can merge when passes.

Issue #312, opened 2019-04-03
OntologyFactory should handle propertyChainAxioms


dustine32:
It would be nice if I could use propertyChainAxioms in certain ontologies (e.g. RO) to connect certain terms. Example use case is deriving "causally upstream of or within, positive effect" (RO:0004047) from "acts upstream of or within, positive effect" (RO:0004032). In the [RO owl](http://purl.obolibrary.org/obo/ro.owl):
```
<owl:ObjectProperty rdf:about="http://purl.obolibrary.org/obo/RO_0004032">
        <rdfs:subPropertyOf rdf:resource="http://purl.obolibrary.org/obo/RO_0002264"/>
        <owl:propertyChainAxiom rdf:parseType="Collection">
            <rdf:Description rdf:about="http://purl.obolibrary.org/obo/RO_0002327"/>
            <rdf:Description rdf:about="http://purl.obolibrary.org/obo/RO_0004047"/>
        </owl:propertyChainAxiom>
```
This connection looks to be in the `owl:propertyChainAxiom` array but it's not coming through to the RO:0004032 node:
```
>>> onto.node("RO:0004032")
{  
   'id':'http://purl.obolibrary.org/obo/RO_0004032',
   'meta':{  
      'basicPropertyValues':[  
         {  
            'pred':'OIO:created_by',
            'val':'cjm'
         },
         {  
            'pred':'rdfs:seeAlso',
            'val':'http://wiki.geneontology.org/index.php/Acts_upstream_of_or_within,_positive_effect'
         },
         {  
            'pred':'RO:0004049',
            'val':'RO:0002264'
         },
         {  
            'pred':'OIO:creation_date',
            'val':'2018-01-26T23:49:30Z'
         }
      ]
   },
   'type':'PROPERTY',
   'lbl':'acts upstream of or within, positive effect',
   'label':'acts upstream of or within, positive effect'
}
```
Currently not sure how this should be stored in the node data structure but I'll try seeing what works for me.

From @cmungall on gitter:
> note in the json the chain is not in the node but in a separate section under”graphs”, "propertyChainAxioms" : [ {...

cmungall:
plan:

new class PropertyChainAxiom, list of these would be linked from the ontology object, structurally the same as the json

so we would do:

```
pcas = ont.get_property_chain_axioms(relation=rnode.id)
```

dustine32:
[...]

PR #313, opened 2019-04-03
Added retrieval of propertyChainAxioms from an obograph file for issu…

dustine32:
…e #312


dustine32:
thanks! I tried to flex my brain on some of the docstrings so @cmungall feel free to correct them if they're wrong.

PR #314, opened 2019-04-04
changing gorule 16 to a filter by changing fail mode to hard

dougli1sqrd:


dougli1sqrd:
For https://github.com/geneontology/go-site/issues/955

PR #319, opened 2019-05-01
adding some log output and turning filter object into list

dougli1sqrd:


dougli1sqrd:
This fixes https://github.com/geneontology/go-site/issues/1061

kltm:
So many coercion rules; what's an object to do?

dougli1sqrd:
It's only a fixed length of two at the moment. This is setup to allow us to add any number extra sources to merge into a file.

Issue #321, opened 2019-05-02
cachier is messing up the pipeline - lets make this optional


dougli1sqrd:
cachier caching mechanism in the ontol_factory is causing issues left and right. Let's make it so cachier doesn't harm the pipeline any longer.

kltm:
Error in pipeline as:
```
[2019-05-02T09:51:20.895Z] Traceback (most recent call last):
[2019-05-02T09:51:20.895Z]   File "/var/lib/jenkins/workspace/eontology_pipeline
_snapshot-OLCOSBORX7TUJKUSVDZQNZDGXCRFSUQLNESYCP3R63U6FLW5DJ2A/go-site/pipeline/
mypyenv/lib/python3.5/site-packages/portalocker/portalocker.py", line 137, in lock
[2019-05-02T09:51:20.895Z]     fcntl.flock(file_.fileno(), flags)
[2019-05-02T09:51:20.895Z] BlockingIOError: [Errno 11] Resource temporarily unavailable
[2019-05-02T09:51:20.895Z] 
[2019-05-02T09:51:20.895Z] During handling of the above exception, another exception occurred:
[2019-05-02T09:51:20.895Z] 
[2019-05-02T09:51:20.895Z] Traceback (most recent call last):
[2019-05-02T09:51:20.895Z]   File "/var/lib/jenkins/workspace/eontology_pipeline_snapshot-OLCOSBORX7TUJKUSVDZQNZDGXCRFSUQLNESYCP3R63U6FLW5DJ2A/go-site/pipeline/mypyenv/lib/python3.5/site-packages/portalocker/utils.py", line 131, in acquire
[2019-05-02T09:51:20.895Z]     fh = self._get_lock(fh)
[2019-05-02T09:51:20.895Z]   File "/var/lib/jenkins/workspace/eontology_pipeline_snapshot-OLCOSBORX7TUJKUSVDZQNZDGXCRFSUQLNESYCP3R63U6FLW5DJ2A/go-site/pipeline/mypyenv/lib/python3.5/site-packages/portalocker/utils.py", line 180, in _get_lock
[2019-05-02T09:51:20.895Z]     portalocker.lock(fh, self.flags)
[...]

PR #322, opened 2019-05-03
set of relations used instead of searching the ontology for URIs

dougli1sqrd:
For https://github.com/biolink/ontobio/issues/321

### DIFF
--- a/ontobio/rdfgen/assoc_rdfgen.py
+++ b/ontobio/rdfgen/assoc_rdfgen.py
@@ -3,7 +3,7 @@ from prefixcommons.curie_util import contract_uri, expand_uri, get_prefixes
 from ontobio.vocabulary.relations import OboRO, Evidence
 from ontobio.vocabulary.upper import UpperLevel
 from ontobio.ecomap import EcoMap
-from ontobio.ontol_factory import OntologyFactory
+from ontobio.rdfgen import relations
 from rdflib import Namespace
 from rdflib import BNode
 from rdflib import Literal
@@ -15,6 +15,7 @@ import rdflib
 import logging
 import uuid
 import re
+import json
 
 ro = OboRO()
 evt = Evidence()
@@ -86,9 +87,9 @@ class RdfTransform(object):
         self.ecomap = EcoMap()
         self._emit_header_done = False
         self.uribase = writer.base
-        self.ro = None
         self.ecomap.mappings()
         self.bad_chars_regex = re.compile("[^\.:_\-0-9a-zA-Z]")
+        self.ro_lookup = dict(relations.label_relation_lookup())
 
     def blanknode(self):
         return BNode()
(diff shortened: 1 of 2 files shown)


kltm:
hard-coded list approved by @cmungall

RichardBruskiewich:
@dougli1sqrd @kltm (@deepakunni3 ) I'm not that convinced that issue 321 was resolved.  The 1.10.2 release of ontobio in an NCATS Translator Workflow2 (running under a venv running on Windows 10 inside a PyCharms IDE...) and I've noticed that the @cachier directive wrapping ontol_factory ends up creating Observer() threads and recursing indefinitely into _PickleCore.wait_on_entry_calc(self, key), with the Stack gobbling up all available system memory at:

```
observer.start()
        observer.join(timeout=1.0)
        if observer.isAlive():
            # print('Timedout waiting. Starting again...')
            return self.wait_on_entry_calc(key)
```

I suppose this could be a timeout issue when running within an environment with some significant overhead, but this clearly this is problematic. Do we really need the @cachier here?

dougli1sqrd:
I'm not sure the situation that NCATS Translator is using ontoibio, but the cachier fixes that I implemented were intended to keep them running in the GO data pipeline (GAF processing) context. This change here eliminates usages of cachier via ontology creation (`OntologyFactory().create(...)`) and to my knowledge that is the only place in `assoc_rdfgen.py` that cachier would be invoked. I could be incorrect though. 

In general, other usages of ontobio that do not use the `bin/validate.py` as an entry point should still have cachier turned on by default, so as to not affect other users. The issues we had with cachier were about the cache not being invalidated correctly, and so not unlocking files. Your issue seems to be wholly different. I am not sure why cachier would be causing problems with spinning out Oberver threads as I do not believe we have experienced this here. 

Replacing cachier with a different caching package is a possibility, but something I don't think we currently have an open issue for, and we would have to schedule in the future.

RichardBruskiewich:
Hello Eric,

The code base upon which I'm working is a so-called "workflow 2" of the NCATS Translator project (see *WF2_automation.py* plus the proximal *Modules* set).  The system imports Ontobio as a tool to load an _in memory_ copy of GO, HP (and perhaps other ontology in the future) which, alongside GO annotations of genes, is used Jaccard similarity analyses.

I'm cc:'ing @deepakunni3 (of Chris Mungall's team) since he's also involved in these efforts.

I'm not sure why, but this usage of the current 1.10.2 release of Ontobio has the very peculiar behaviour (at least, when run on my Windows 10 machine from inside the PyCharms IDE) of the *cachier* tagged functions for *assoc_factory* and *ontol_factory*  recursing deeply into the following cachier module *pickle_core.py* code snippet: 

```        observer = Observer()
        event_handler.inject_observer(observer)
        observer.schedule(
            event_handler,
            path=EXPANDED_CACHIER_DIR,
            recursive=True
        )
        observer.start()
        observer.join(timeout=1.0)
        if observer.isAlive():
            # print('Timedout waiting. Starting again...')
            return self.wait_on_entry_calc(key)``` 

resulting in an indefinite series of new *Observer()* threads being started, along with stack memory usage progressively consuming up all the available free memory of the machine, before Windows 10 starts to thrash into swapping with hard disk pages (and slowing the operating system to a syrupy crawl!).

For our usage of Ontobio, we mainly use it to access the GO or HP ontology catalogs once from the remote server, for runtime hosting inside a huge _in memory_ data structure inside our application.
 
Our script is expected to either be run standalone on someone's machine or on an NCATS server process dedicated to running the workflow (e.g. workflow #2).  We suspect that caching remote access to the data source makes little sense in this context. Thus, if *cachier* function is sensitive to this particular execution context (and has this odd deep recursion / stack memory consumption behaviour), then it is perhaps counterproductive to our use case.

I have (yesterday) applied an initial patch to a local fork of Ontobio, which adds a new _config.yaml_ boolean configuration parameter called "ignore_cache" which would be set to "True" for our software but perhaps "False" otherwise. I have then propagated the "ignore_cache" to the invocation points of various \@cachier tagged functions. 

Alas, I'm only really just seriously starting to test the behaviour of this modified version this morning to see if the performance issues we observed above are mitigated. 

I was not completely sure precisely what \@cachier tagged functions really need to be disabled in this context, so I may go back and later be more selective in the use of *ignore_cache* as required.

PR #324, opened 2019-05-06
adding gorule6 qc rule to the suite of rules in the parser

dougli1sqrd:


dougli1sqrd:
This is already done, this PR is not needed.

PR #326, opened 2019-05-08
references can only have one id per id-space

dougli1sqrd:
For https://github.com/geneontology/go-site/issues/1063

For each reference in a gaf line, if there is more than one reference ID with the same ID prefix (ID space) then we throw a warning and gets included in the report under gorule 1.


dougli1sqrd:
@cmungall @kltm I want to confirm that this check should skip (filter) lines that violate the principle of having more than one ID in an ID space per line. OR we just throw a warning and NOT skip (filter)? The PR as written currently does do a filter.

kltm:
As written, https://github.com/geneontology/go-site/blob/master/metadata/rules/gorule-0000001.md states that it is a hard rule, and thus filters when running into issues. If it only warned, it seems like it would be a separate rule. This is rule 0000001 as it is considered a "basic" format error. @pgaudet , does this all sound good and correct to you?

pgaudet:
I would like to have it as a Warning so we can evaluate how many annotations fail. 

When that number is reasonably low we can move it to a Filter. 

(I dont know what the committed code does, so I will let you decide to merge if this is in accordance with this).

dougli1sqrd:
@pgaudet While gorule 1 is usually a filter/hard violation, I have made this particular syntax check a warning/report.

dougli1sqrd:
In preference for https://github.com/biolink/ontobio/pull/469, I am closing this PR.

PR #327, opened 2019-05-09
removing group specific reference IDs

dougli1sqrd:
I need to write a couple tests for this new function, but then will be actual ready. Take a look in the meantime.


kltm:
This seems like it's likely a fairly straight forward implementation of 33

PR #328, opened 2019-05-14
rule 33 impl only reports, does not repair

dougli1sqrd:
Per the changes to gorule 33 to make it report only, this makes no changes to the references.


dougli1sqrd:
Seeing that the gorule has been merged, I will merge this as well.

PR #342, opened 2019-05-21
Filter for None before calling join in golr_query

deepakunni3:
This bug was causing some of the Golr queries to fail

### DIFF
--- a/ontobio/golr/golr_query.py
+++ b/ontobio/golr/golr_query.py
@@ -425,7 +425,7 @@ class GolrSearchQuery(GolrAbstractQuery):
         params = {
             'q': '{0} "{0}"'.format(self.term),
             "qt": "standard",
-            'fl': ",".join(select_fields),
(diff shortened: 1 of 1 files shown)


kshefchek:
Is this something that would be better fixed upstream - do we know where the nulls are coming from?

deepakunni3:
The nulls are showing up when we map the GOlr fields with AmiGO GOlr fields.
The field mapping contains None, which then finds its way into `select_fields`

kshefchek:
makes sense, I haven't taken a close look but my intuition is that we could filter these out when creating/appending items to the list, or adding this filter right after all items have been added

deepakunni3:
Yup, makes sense. I think the entirety of golr_query is due a sanity check.
Because we should handle empty lists and `None` values properly before sending it to Solr.
There are bugs popping up when certain fields are None and they are passed through.

For example, 
- `fq=subject_taxon_closure:""` yields no result
- `fq=subject_taxon_closure:"None"` yields no result

PR #343, opened 2019-05-24
Implementing relation-aware ancestors by DFS locally, no longer using nx, fixes #335

### LINKED ISSUE #335: Improve performance of ontol.get_ancestors when relations are passed
Currently when we pass in a list of relations to an ontol object, a new networkx graph is created, with that set of relations. This incurs a large performance penalty at the outset while the subgraph is constructed. The client can avoid this by building these subgraphs in advance but this is awkward.

We should just implement relation-aware ancestors ourselves. I had originally thought that nx would provide a fast C implementation but on cursory examination that appears not to be the case.

cmungall:


deepakunni3:
@cmungall I don't see any difference in performance. Its on par with `networkx.ancestors()`

cmungall:
@dougli1sqrd can you review the code and confirm this causes no change in pipeline outputs

kltm:
To verify pipeline outputs, we'll need a dummy version increment and train the pipeline onto that "release".
To make this process easier, we may want to figure out a way of being able to separate out ontobio.

dougli1sqrd:
The pipeline only uses the `ancestors` call to compute the aspect field of an annotation. This honestly will probably be shortly replaced, since the classes in the ontology knows which namespace it's in (and the current method is slow comparatively). As long as this still finds `GO:0008150`, `GO:0005575`, `GO:0003674` for any leaf node everything should be the same in the meantime.

dougli1sqrd:
I can verify this is quite fast compared to regular version!

dougli1sqrd:
Would we also want to do this for descendants as well?

PR #347, opened 2019-06-06
WIP initial draft of enables o has part annotation inference

dougli1sqrd:
python bin/materialize.py infer --ontology go-ontology.json --gaf tmgi.gaf --target mgi_enables_inferred.gaf


dougli1sqrd:
@cmungall what did you suggest I rename the `ancestors()` function to?

PR #351, opened 2019-06-21
gorule 37 implementation in ontobio qc

dougli1sqrd:


kltm:
I trust this works--we'll see in snapshot.

kltm:
Wait...it's gone now...

PR #354, opened 2019-07-02
add option to not download if file already exists

dougli1sqrd:


kltm:
Will trust in @dougli1sqrd

Issue #358, opened 2019-08-07
Cachier is not thread safe and causes problems in some environments


deepakunni3:
Most recently, the GO team has had cachier crash in a multi-process environment when using the rdfgen command line.

In general, cachier does not work well in many environments and should be replaced with something safer.

Noting that #321 is to make cachier optional. However, this ticket proposed closing that and removing it as it is not safe.

Tagging @dougli1sqrd @kltm (ghost wrote this issue) @cmungall

deepakunni3:
Based on a quick discussion with @kltm and @cmungall , the decision is to remove cachier completely as the benefits are not worth the hassle.
We will have a more deliberate caching mechanism, one where user has more control, and the caching behavior is explicitly defined.

dougli1sqrd:
The suggestion is to remove it, and then later add a different caching mechanism?

kltm:
That is what I got from the conversation with @cmungall and @deepakunni3 .

dougli1sqrd:
Cool cool. I agree.

cmungall:
yes. More explicit caching, allow command line user to save ontology as
pickle file, allow pickled ontology to be loaded using normal ontology load
mechanisms

kshefchek:
Caching is pretty important for calls to owlsim, here are some benchmarks comparing no caching vs diskcache:
[cache-comparison.txt](https://github.com/biolink/ontobio/files/3511463/cache-comparison.txt)

cmungall:
OK we can restore it for owlsim for now (I assume you don't need to cache
the ontology loading part, just the web service calls). This won't affect
the GO pipeline.

I don't know where that leaves us with the Translator workflows

PR #359, opened 2019-08-07
Remove cachier

deepakunni3:
This PR removes the usage of Cachier from Ontobio.
An alternate caching mechanism will be introduced as a subsequent PR.

Refer to #358 for more on this discussion.

### DIFF
--- a/ontobio/sim/api/owlsim2.py
+++ b/ontobio/sim/api/owlsim2.py
@@ -10,20 +10,17 @@ from ontobio.util.scigraph_util import get_nodes_from_ids, get_id_type_map, get_
 from typing import List, Optional, Dict, Tuple, Union
 from json.decoder import JSONDecodeError
 import logging
-from cachier import cachier
-import datetime
 import requests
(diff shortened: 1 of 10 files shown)


deepakunni3:
@kltm thanks for double checking. I found references to cachier in some notebooks (not a concern), documentation (fixed) and setup.py (fixed).

ontobio/sim/api/owlsim2.py
```diff
@@ -10,20 +10,17 @@
...
 from json.decoder import JSONDecodeError
 import logging
-from cachier import cachier
-import datetime
 import requests
 
 """
 Functions that directly access the owlsim rest API are kept
 outside the class to utilize cachier
+TODO: this will have to be revisited.
```
kltm: Fair enough.

deepakunni3:
The references to cachier in the notebooks are basically just cachier warnings that have been made persistent as part of the cell output. It does not directly call cachier.

PR #360, opened 2019-08-08
Use `fastobo` to load ontologies from OBO files

althonos:
Hi !

This PR (finally) adds a native OBO parser using the [`fastobo`](https://pypi.org/project/fastobo/) package. This is very much WIP since there are no actual semantics for OBO to OBO Graphs translation, and the translation is in early development.

I added it in a way `fastobo` becomes a mandatory dependency, which is not optimal on platforms other than Linux and OSX x86-64 since it requires a Rust toolchain to be available on the target machine; I could make it optional as well if you think this is better.

### DIFF
--- a/ontobio/ontol_factory.py
+++ b/ontobio/ontol_factory.py
@@ -13,8 +13,10 @@ from ontobio.sparql.sparql_ontology import EagerRemoteSparqlOntology
 import os
 import subprocess
 import hashlib
+import fastobo
 from cachier import cachier
 import datetime
+import tempfile
 
 SHELF_LIFE = datetime.timedelta(days=3)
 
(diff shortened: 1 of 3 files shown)


althonos:
> So this requires rust on the machine that is running fastobo?
Not at runtime, you don't need Rust once the [`fastobo`](https://pypi.org/project/fastobo) package is installed.

> Does pip install this locally in a virtual environment? 
This is like any Python library, you can use fancy stuff like `venv` or just do a site install with `pip`.

> Like, will someone have to separately install rust before running this?
If they are running an x86_64 Linux or OSX, no, because I provide pre-built versions. If they are not, they need a Rust toolchain at compile-time, but once the module is compiled the toolchain can be uninstalled.

PR #365, opened 2019-08-16
gorule 15, interacting taxon check

dougli1sqrd:


kltm:
All seems cool to me.

PR #375, opened 2019-09-19
implementing gorule 43

dougli1sqrd:


kltm:
nothing seems off to me

PR #377, opened 2019-09-23
adds a proper shebang to common ontobio scripts

dougli1sqrd:


kltm:
This does not immediately affect me/GO. Perhaps @deepakunni3 might have a thought?

dougli1sqrd:
Shouldn't affect anything but the pipeline. Just throwing your way for some eyes.

kltm:
Seems unlikely to break things.

deepakunni3:
I know this is after the fact, but it doesn't affect anything. Much more neater this way.

Issue #388, opened 2019-10-18
Parsing a GAF file fails due to a bug in implementation of gorule-0000046


deepakunni3:
The latest release (and master branch) yields the following error when trying to parse a GAF file:

```py
>>> p = GafParser()
>>> url = "http://current.geneontology.org/annotations/goa_human.gaf.gz"
>>> assocs = p.parse(url)
```

Stack trace:
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/deepak.unni3/GIT/ontobio-master/ontobio/io/assocparser.py", line 307, in parse
    a = list(associations)
  File "/Users/deepak.unni3/GIT/ontobio-master/ontobio/io/assocparser.py", line 320, in association_generator
    parsed_result = self.parse_line(line)
  File "/Users/deepak.unni3/GIT/ontobio-master/ontobio/io/gafparser.py", line 146, in parse_line
    go_rule_results = qc.test_go_rules(list(vals), self.config)
  File "/Users/deepak.unni3/GIT/ontobio-master/ontobio/io/qc.py", line 511, in test_go_rules
    result = rule.value.run_test(active_annotation, config)
  File "/Users/deepak.unni3/GIT/ontobio-master/ontobio/io/qc.py", line 64, in run_test
    result = self.test(annotation, config)
  File "/Users/deepak.unni3/GIT/ontobio-master/ontobio/io/qc.py", line 459, in test
    if goterm in self.self_binding_terms:
TypeError: argument of type 'NoneType' is not iterable
```

It looks like the `TypeError` originates from `GoRule46` in `qc.py`: https://github.com/biolink/ontobio/blob/master/ontobio/io/qc.py#L443

deepakunni3:
As a side note, our test suite seems to pass successfully. Curious to know why this wasn't captured earlier and if there is a test we can add that can catch this during development.

dougli1sqrd:
Thanks for catching this Deepak! First thing I'm noticing is that you are parsing a url. Do you normally use gafparser in this manner? That's definitely not a use case that I use, so that's a pathway I'm far less familiar with. I'll definitely take a look though! And yes, we should definitely have a test that covers this if we can.

dougli1sqrd:
I see what's going on. Yeah, looks like we just need to set that variable in the case that we do not have an ontology passed in. I wrote a test and made it pass. We now just set the full list of self-interacting terms to only the known root terms if we do not have an ontology.

PR #390, opened 2019-10-21
fixing bug with gorule46 and passing in no ontology

dougli1sqrd:
https://github.com/biolink/ontobio/issues/388

@deepakunni3 Will we need a release for this? Or is master sufficient for now?


deepakunni3:
@dougli1sqrd Looks good!  👍 

A release would be a better idea since its been a while since the last release.

PR #391, opened 2019-10-24
Go site issue 970 example runner

dougli1sqrd:
This is for https://github.com/geneontology/go-site/issues/970

This creates a new command for the validate.py script: `rule`. You give it a path to the go-site metadata and an ontology and it will validate the `examples` given in go rules. We made the schema for this a while back. One can add exemplar data that is supposed to `PASS`, `FAIL`, or be `REPAIR`ed. `rule` will create a parser, parse the example data, and compare with what it was expected to get. An example is validated if it matches the state that the example was in. As in: if a gaf line is expected to fail a certain rule, and the parser successfully finds an error for that rule in the given example line, then that example is valid. 

Any invalid example will make the script exit with `1`. This should be used in ontobio travis to validate that the implementation of the rules are correct. 

Optionally a report JSON file will be written out. This could be run in the pipeline and be viewable as a product.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,7 +25,8 @@ foo:
 # only run local tests
 travis_test:
 	pytest tests/test_*local*.py tests/test_*parse*.py tests/test*writer*.py tests/test_qc.py \
-	       tests/test_rdfgen.py tests/test_phenosim_engine.py tests/unit/ tests/test_ontol.py
+	       tests/test_rdfgen.py tests/test_phenosim_engine.py tests/unit/ tests/test_ontol.py \
+		   tests/test_validation_rules.py
 
 cleandist:
 	rm dist/* || true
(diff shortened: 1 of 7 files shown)


kltm:
IIRC, originated from Suzanna and carried forward from there.
A quick scan turns up nothing to my eye, although I'm curious about all the "normalize" commands.
I'd vote merge if there are no objections as it seems to be isolated to mostly go-ish sub-areas.

dougli1sqrd:
Gaf lines are allowed to have between 15 and 17 columns, and is equivalent to 17 columns with the last two as empty strings. This creates some comparison difficulties where an "expected" might only have 15 columns, but the parsed input will always have 17 columns since we "normalize" it to 17 columns.

PR #397, opened 2019-11-20
removing validation for gorule-000033 since it was deprecated

dougli1sqrd:
See https://github.com/geneontology/go-site/pull/1197
And https://github.com/geneontology/go-site/issues/1082#issuecomment-556051891

gorule-0000033 will be deprecated, and so we will stop validating on that rule.


dougli1sqrd:
gorule 33 is now deprecated, so merging this to reflect that.

Issue #403, opened 2019-12-04
GpadParser `parse_line` should minimally and immediately create a GoAssociation object from the line


dougli1sqrd:
In order to allow processing GoAssociations in GO Rules the various parsers need to create GoAssociation objects out of incoming GPAD lines.

This is a as-minimal-as-possible change for now. We should do a slightly larger re-ordering of things once we're more aligned, but this is incremental for now.

To do:

- [x] Do we need to change ParseResult, or add a lightweight Header object?
- [x] Any problem parsing the line into a GoAssociation is essentially a gorule-0000001 failure
- [x] I don't think this requires us to change the output dictionary that is currently representing associations yet. We're only trying to plug this in to the qc GO Rules engine in ontobio
- [x] plug in the GO Rules runner into the gpad parser, like in the gaf parser

PR #412, opened 2019-12-12
Issue 295 qc refactor

dougli1sqrd:
This is the initial qc refactor, so that the Rules engine is now based on the GoAssociation data model. This will allow us to make a similar pattern for gpad parser. 

This creates a minimal shunt from gafparser into the rules, where we convert GAF lines, List of String --> GoAssociation. 

In https://github.com/biolink/ontobio/issues/402 and https://github.com/biolink/ontobio/issues/403 we can make a slightly more sophisticated translation from TSV-like data into GoAssociation data. In https://github.com/biolink/ontobio/issues/403 in particular we will also add the GO Rules to GPAD parser in the first place.


kltm:
Nice to see object refs instead of array refs. Trusting this passes tests. Large enough that it should wait for another review.

dustine32:
@dougli1sqrd Just noting: when coding #403 we'll need to calculate the aspect for every GPAD line (likely at parsing time) in order to use the GO rules since there's no aspect col in the GPAD spec.

dougli1sqrd:
Okay going to merge this.

Issue #416, opened 2020-01-08
Qualifier to relations CURIE crashes when there isn't a corresponding relation for the label


dougli1sqrd:
In building Associations for GAF/GPAD processing, there are points where we convert between labels of relations to CURIEs of those relations. If a relation doesn't exist for a label, then previously ontobio would just crash.

The lookup should be more robust, and we should report errors when we cannot find relations.

PR #417, opened 2020-01-08
fixing curie relation issue from #416

dougli1sqrd:
Fixing https://github.com/biolink/ontobio/issues/416

### DIFF
--- a/ontobio/io/gafparser.py
+++ b/ontobio/io/gafparser.py
@@ -418,11 +418,15 @@ def to_association(gaf_line: List[str], report=None, group="unknown", dataset="u
             conjunction = association.ExtensionConjunctions(extension_units)
             conjunctions.append(conjunction)
     object_extensions = association.ExtensionExpression(conjunctions)
+    looked_up_rel = relations.lookup_label(relation)
+    if looked_up_rel is None:
+        report.error(source_line, assocparser.Report.INVALID_QUALIFIER, relation, "Qualifer must be \"colocalizes_with\" or \"contributes_to\"", taxon=taxon, rule=1)
+        return assocparser.ParseResult(source_line, [], True, report=report)
 
     a = association.GoAssociation(
         source_line="\t".join(gaf_line),
         subject=subject,
-        relation=curie_util.contract_uri(relations.lookup_label(relation))[0],
+        relation=curie_util.contract_uri(looked_up_rel)[0],
         object=object,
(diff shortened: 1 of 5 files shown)


ontobio/io/qc.py
```diff
@@ -511,7 +510,8 @@ def test(self, annotation: association.GoAssociation, config: assocparser.AssocP
     "GoRule42": GoRule42(),
     "GoRule43": GoRule43(),
     "GoRule46": GoRule46(),
-    "GoRule50": GoRule50()
+    "GoRule50": GoRule50(),
+    "GoRule13": GoRule13()
```
kltm: I'd add a comment here for the rationale, so that somebody doesn't switch it back; e.g.:
```python
// WARNING: This rule appears last as we want to clean up the input as much as possible before invoking a rule that depends on an external program. See issue XYZ for some history.
```

ontobio/io/gafparser.py
```diff
@@ -418,11 +418,15 @@ def to_association(gaf_line: List[str], report=None, group="unknown", dataset="u
             conjunction = association.ExtensionConjunctions(extension_units)
             conjunctions.append(conjunction)
     object_extensions = association.ExtensionExpression(conjunctions)
+    looked_up_rel = relations.lookup_label(relation)
+    if looked_up_rel is None:
+        report.error(source_line, assocparser.Report.INVALID_QUALIFIER, relation, "Qualifer must be \"colocalizes_with\" or \"contributes_to\"", taxon=taxon, rule=1)
```
kltm: Or "NOT"?

ontobio/io/qc.py
kltm: Ah, I see it up there at the top now. Honestly, would you look at the top before switching it back?

ontobio/io/gafparser.py
dougli1sqrd: I didn't include NOT since NOT is filtered out of the "relation" like qualifiers. And so NOT is not examined in the same way as the other qualifiers are here. If that is confusing though, I can say NOT also.

ontobio/io/qc.py
dougli1sqrd: I put it at the top since it's sort of a python idiom to document classes and functions from the top.

kltm:
Talked with @dougli1sqrd , he'll merge when he's happy.

PR #418, opened 2020-01-16
Hackathon 2020 rules

dougli1sqrd:
This is the work we accomplished over the hackathon: adding gorule 57, fixed some gpad parsing, added tests.


kltm:
From discussions over the last two days, I believe that this is "master safe".

kltm:
@dougli1sqrd Retraction: the metadata shouldn't be in here for the `master`, right?

dougli1sqrd:
I think you're just seeing a resource file for testing of the group specific filters.

PR #421, opened 2020-02-12
setup logging configurations and start out some heavier duty logging …

dougli1sqrd:
…around the report file


kltm:
The Promised Logging.

PR #422, opened 2020-02-12
making logging setup more robust

dougli1sqrd:
For some reason the logging configuration I was using before failed after all the merges. Here is a more robust version. I'm going to just merge this through once travis passes


kltm:
geneontology/go-site#1380

PR #423, opened 2020-02-13
adding some more logging in validate.py, removed verbose Expand log message

dougli1sqrd:
@valearna, @sierra-moxon, in assoc_rdfgen.py line 104 there is a log message for when we expand a CURIE into a URI.

Do you need this in your logs? I will be commenting this out in this PR as it is spamming our logs. Let me know if that messes you up!

Thanks,
Eric


valearna:
I don't think we need it in our logs, right @sierra-moxon? It is probably a log line from an older ontobio version (the one I originally forked).

sierra-moxon:
Nope don’t need it, makes sense.

PR #426, opened 2020-02-18
Go site issue 1253 missing gorule 13

dougli1sqrd:
This is a fix for https://github.com/geneontology/go-site/issues/1253

Essentially, this adds a flag to the function that produces validated GAFs. If it's a mixin (like paint_pombase for pombase) then the `mixin` flag is set to `True`.

With mixin set to true, it doesn't generate the same report. This prevents running pombase from producing a `paint_pombase.report.md` which collides with running paint and it's own `paint_pombase.report.md`.


kltm:
Assuming this is the overlapping "namespace" fix.

PR #428, opened 2020-02-27
extending rules for gorule 13

dougli1sqrd:
For https://github.com/geneontology/go-site/issues/882.

### DIFF
--- a/ontobio/io/qc.py
+++ b/ontobio/io/qc.py
@@ -190,21 +190,34 @@ class GoRule11(GoRule):
 class GoRule13(GoRule):
 
     def __init__(self):
-        super().__init__("GORULE:0000013", "Taxon-appropriate annotation check", FailMode.SOFT)
+        super().__init__("GORULE:0000013", "Taxon-appropriate annotation check", FailMode.HARD)
+        self.non_experimental_evidence = ["ECO:0000318", "ECO:0000320", "ECO:0000321", "ECO:0000305", "ECO:0000247", "ECO:0000255", "ECO:0000266", "ECO:0000250", "ECO:0000303", "ECO:0000245", "ECO:0000304", "ECO:0000307"]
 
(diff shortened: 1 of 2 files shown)


ontobio/io/qc.py
```diff
@@ -190,21 +190,34 @@ def test(self, annotation: association.GoAssociation, config: assocparser.AssocP
 class GoRule13(GoRule):
 
     def __init__(self):
-        super().__init__("GORULE:0000013", "Taxon-appropriate annotation check", FailMode.SOFT)
+        super().__init__("GORULE:0000013", "Taxon-appropriate annotation check", FailMode.HARD)
+        self.non_experimental_evidence = ["ECO:0000318", "ECO:0000320", "ECO:0000321", "ECO:0000305", "ECO:0000247", "ECO:0000255", "ECO:0000266", "ECO:0000250", "ECO:0000303", "ECO:0000245", "ECO:0000304", "ECO:0000307"]
```
kltm: This should be a more generally available and visible variable.

dougli1sqrd: Fair enough. Practically though, this could be difficult. The current state of affairs is that ontobio only has access to GO. The rule itself refers to these (by evidence code) specifically as well. So any updates to the ontology or mappings, etc currently need to be propagated through the rule and into the implementation, which is unfortunate.

An improvement could be that the rule would state a process to arrive at this list from the wording "non experimental evidence". Then at least if ontobio had ECO at some point it could do that ontology computation.

Then just as an aside, this is a pattern that the rules in `qc.py` use, so it's consistent at this time. Things that rules state directly (example: "use these evidence IDs") are usually put in the constructor `__init__` like this.

kltm:
Again, this is for future, but the concept of an experimental and non-experimental grouping comes up a lot and is of general interest. We'll get there eventually.

PR #429, opened 2020-03-03
all validations now produce reports like before v1.15.6

dougli1sqrd:


kltm:
While a "large" change, from discussion this is just a reversion. Feel free to take if you feel this is unlikely to be problematic.

PR #430, opened 2020-03-05
Issue 427 rulesets

dougli1sqrd:
This adds a command line option `--rule-context/-T` to ontobio validate that takes a "rule context".

In the GO Rules, we allow a _tag_ of `context-import` to indicate that the rule should be run during an import pipeline run, but not in an ordinary run.

To specify that a the validation should run the import rule set, the option `-T import` will tell ontobio to run any rule that has the `context-import` tag. When this is not set, rules that have the tag will be skipped. However, they will still appear in the reports as having no errors.


dougli1sqrd:
Merging.

PR #440, opened 2020-04-01
report error if qualifiers are wrong in any way

dougli1sqrd:
…, filter wrong qualifiers

For https://github.com/geneontology/helpdesk/issues/252

This fixes a bug where malformed qualifiers and wrong qualifiers not allowed were all allowed through the parser with no issue.

Qualifiers with spaces will be repaired with underscores.


kltm:
Alas, I trust that this works, but let's add this to the let's-do-after-`release` pile.

dougli1sqrd:
Okay. I'll update the implementation then. Once this is merged I can release ontobio, and update go-site.

PR #441, opened 2020-04-02
adding rule 13 test case for IEA, adding IEA to non exp eco ids

dougli1sqrd:
For https://github.com/geneontology/go-site/issues/1444


kltm:
Assuming works as advertized.

PR #443, opened 2020-04-10
Adjust window expansion for IEAs in GAF QC

kltm:
Work on geneontology/go-site#1456

### DIFF
--- a/ontobio/io/assocparser.py
+++ b/ontobio/io/assocparser.py
@@ -533,27 +533,6 @@ class AssocParser(object):
 
         return id
 
-    def _normalize_gaf_date(self, date, line: SplitLine):
-        if date is None or date == "":
-            self.report.warning(line.line, Report.INVALID_DATE, date, "GORULE:0000001: empty",
-                taxon=line.taxon, rule=1)
-            return date
-
-        # We check int(date)
-        if len(date) == 8 and date.isdigit():
-            d = datetime.datetime(int(date[0:4]), int(date[4:6]), int(date[6:8]), 0, 0, 0, 0)
-        else:
-            self.report.warning(line.line, Report.INVALID_DATE, date, "GORULE:0000001: Date field must be YYYYMMDD, got: {}".format(date),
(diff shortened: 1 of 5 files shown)


kltm:
@dougli1sqrd Okay, stuck at parsing a date in the tests...which confuses me.

dougli1sqrd:
I just had a brief look, and it looks like you've stumbled upon a brittleness in the date parsing for GoAssociation. This should be pretty straightforward to fix though.

Right now we just pass the date through directly from the gaf line into the date field of the GoAssociation. Not sure how this passed the date tests that are failing here before, honestly. But either way was clearly not correct. 

What's happening is the tests have example bad dates, like 2015-03-05, and dashes are not correct. This gets directly passed through into a GoAssociation, gorules on run on it, and then the rest of the gaf parser proceeds. One of the things the gaf parser does though is validate and normalize the date, but this doesn't happen in the gorules directly, but *after* (but parse failures of the date are marked as gorule 1 failures as it should be). 

Somehow the old code wasn't producing a parse error even though it's doing similar things as code here, gets past the gorules, goes on to the date normalization where (in this case) the date will be fixed to 20150305, emitting a warning at the same time.

To correct all this, this date normalizing code should be inside the code that produces a GoAssocation.

dougli1sqrd:
Just for sanity's sake, this line in the old code was the thing preventing the failure:
```
fails = (evidence == iea and now - datetime.datetime(int(date[0:4]), int(date[4:6]), int(date[6:8]), 0, 0, 0, 0) > time_compare_delta)
```
and specifically the `evidence == iea` bit. This was preventing the test failures because in the test line the evidence is ISO, so it doesn't even do the date comparison or parsing.

kltm:
Good to know that the short circuit was the reason. Sanity is good.

kltm:
@dougli1sqrd Ready for you to...I guess make a release and add back to go-site?

dougli1sqrd:
Taking now!

Issue #446, opened 2020-04-30
Ontobio should support gaf 2.2 - Qualifiers update


dougli1sqrd:
GAF 2.2 has an updated spec on how Qualifiers are handled.

See https://github.com/geneontology/geneontology.github.io/blob/issue-go-annotation-2917-gaf-2_2-doc/_docs/go-annotation-file-gaf-format-22.md#qualifier-column-4 for the new spec.

ontobio will 
- look for a version in the header now
- save it off for reference when processing lines
- determine separate qualifier logic per version
- have a default version mechanism

dougli1sqrd:
This is complete and in ontobio since v1.18.0

PR #447, opened 2020-04-30
ontobio now respects gaf 2.2 and parses qualifiers dependent on versions

dougli1sqrd:


kltm:
Is this it?! Sweet!
I'm assuming that one test with qualifiers and no version barfs, yes?
Otherwise, this is terrific...and out of an abundance of caution let's merge it after the next release.

dougli1sqrd:
The test with no version will just default to gaf 2.1. Essentially anything that isnt' gaf 2.2 will default to 2.1, since 2.1 is a superset of all other previous versions

dougli1sqrd:
For https://github.com/biolink/ontobio/issues/446

dougli1sqrd:
@kltm So the gaf writer now skips lines that don't comform, writing an error message.

kltm:
Okay, great. As from the email, this is not expected to be an issue, but we'll want to make sure that there is documentation for other users. Would documentation go with this PR, or elsewhere?

dougli1sqrd:
I'm not sure. Maybe in any existing pipeline documentation we could talk about gaf 2.2 and now what the parser does.

kltm:
I'm more thinking for somebody trying to use ontobio--the fact that annotation would be "lost" in conversion should just be mentioned somewhere public.

dougli1sqrd:
ah okay, sure. 

We don't actually have much documentation for ontobio as far as I know. I think there is a website. 

But I could write a docstring in the GafWriter code directly that explains how that works.

kltm:
If there's nothing else, for now that would be the place to start--the only place where somebody could go.

dougli1sqrd:
Updated documentation for GafWriter explaining the version conditions for which a line could be skipped.

Issue #448, opened 2020-05-12
config schema validation returns 'not a valid URL' in docker-compose


kshefchek:
during config schema validation: https://github.com/biolink/ontobio/blob/c70b08/ontobio/config.py#L236

I get an error for URLs from docker-compose, eg
http://solr:8983/solr/golr

Looks like we need to add the flag require_tld=False, https://marshmallow.readthedocs.io/en/3.0/api_reference.html#marshmallow.fields.Url

PR #449, opened 2020-05-12
add require_tld=False to EndpointSchema

kshefchek:
fixes https://github.com/biolink/ontobio/issues/448


kshefchek:
this required increasing 3.0.0b3 to 3.0.0b11 which in turn required some small changes to the load config function

PR #464, opened 2020-07-10
Update README-developers.md

deepakunni3:
Just clarifying the steps needed to make a new Ontobio release.

### DIFF
--- a/README-developers.md
+++ b/README-developers.md
@@ -7,15 +7,29 @@ brew or apt-get or yum.
 This section is only relevant for project maintainers.
 To create a new release, do the following:
 
-1. Bump the `__version__` in [`ontobio/__init__.py`](ontobio/__init__.py).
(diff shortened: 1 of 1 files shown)


dougli1sqrd:
This looks good! Just out of curiosity though, why would the local checked out version ever be different than what's on pypi?

deepakunni3:
Because it was. That's why I ended up skipping `1.19.1`.
The local checkout version was `1.19.1` and I assumed it was already released, and bumped it again to `1.19.2` when in reality PyPI was at `1.19.0`

dougli1sqrd:
oh weird, did I mess something up? Master should have had the right version.

deepakunni3:
Looking back at the commits, it looks like Dustin bumped the version as part of a PR: https://github.com/biolink/ontobio/commit/a15975e76ad713e7650081c36253c9d9e08b5627#diff-5ce3cf331b5fc020557daa5b57bfcea8

I didn't realize that when I merged the PR. So when I bumped, it ended up to be a double bump.

It's all good 👍

Issue #473, opened 2020-09-15
GoRule57 filter lines by provided_by


dustine32:
For the MOD imports project we need to port over the ability to filter for only lines with particular Provided_by values (`MGI` for the MGI GPAD, `WB` for the WB GPAD). Basically, we want to filter out all lines that are not from the MOD (e.g. UniProt, ARUK-UCL, SynGO). This was originally implemented for gocamgen: https://github.com/geneontology/gocamgen/issues/30

Since we already have a `filter_out` section in the go-site dataset YAML files for each group, we can specify what to filter _for_  with a new `filter_for` or `filter_in` section?
```
filter_for:
  provided_by:
    - MGI
```
Hopefully, this doesn't over-complicate things.

Tagging @dougli1sqrd

PR #474, opened 2020-09-15
GoRule 57 update and test for issue #473

dustine32:
@dougli1sqrd I'll probably need to update the [mgi.yaml](https://github.com/geneontology/go-site/blob/master/metadata/datasets/mgi.yaml) and [wb.yaml](https://github.com/geneontology/go-site/blob/master/metadata/datasets/wb.yaml) files in [go-site](https://github.com/geneontology/go-site/tree/master/metadata/datasets) as well, right? Example:
```
filter_for:
  provided_by:
    - MGI
```

### DIFF
--- a/ontobio/io/qc.py
+++ b/ontobio/io/qc.py
@@ -576,6 +576,10 @@ class GoRule57(GoRule):
         if annotation.evidence.type in evidence_codes:
             return self._result(False)
 
+        provided_bys = config.group_metadata.get("filter_for", {}).get("provided_by", [])
+        if provided_bys and str(annotation.provided_by) not in provided_bys:
+            return self._result(False)
+
         evidences_references = config.group_metadata.get("filter_out", {}).get("evidence_reference", [])
         for er in evidences_references:
             evidence_code = er["evidence"]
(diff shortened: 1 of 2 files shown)


kltm:
What's the difference between `filter_for` and `filter_out`? We'll also need to add this to the schema (where we also probably want to document).

dustine32:
@kltm Oh yeah, I forgot to update the [datasets.schema.yaml](https://github.com/geneontology/go-site/blob/master/metadata/datasets.schema.yaml), thanks! I'll make a ticket and PR in `go-site` for this.

Existing `filter_out` defines annotation values that would trigger a line to be filtered out, whereas `filter_for` (maybe we should change this to `filter_in`) defines annotation values that a line must have to prevent being filtered out.

The example in issue #473 shows `filter_for` being used to filter the MGI GPAD, keeping only the lines where Provided_by=MGI and excluding all other lines (e.g. UniProt, SynGO).

kltm:
@dustine32 Just as a note, I'm fine with properties that have long\_and\_human\_readable\_names.
That said, whatever works for you, I just want to make sure that people reading the metadata without all of the context have some idea what's going on (at some point in the future).

dustine32:
Linking schema PR: https://github.com/geneontology/go-site/pull/1554

dustine32:
Thinking about changing `filter_for` to `only_allow` as @dougli1sqrd suggested. Still brainstorming though.

PR #479, opened 2020-09-29
Issue 477 curie leaking

dougli1sqrd:
Curies were being leaked from an unstringed taxon Curie object into the report. So when we went to turn the report into JSON it broke since the Curie object doesn't know how to be a JSON object. Now Curies are converted into strings before submitted to the report.


kltm:
Looks good to me--small fix, merge as you want. 
What an HpoaParser?

dougli1sqrd:
Not sure, it's a different TSV format that Chris knows about though.

Issue #432, opened 2020-03-06
Add command to ontobio that will take input annotations (GAF/GPAD) and outputs TTL


dougli1sqrd:
Ontobio needs to have a new command specifically to create graph models (TTL) from annotation lines (GPAD/GAF).

(This is what gocamgen does)

* Probably the command will be in `validate.py`, along side `produce`.
    * Let's discuss what this should be called.
* There should be a module created where most of the work for this can be made in freedom. (perhaps just rdfgen)
    * the contents of `assoc_rdfgen.py`, though should be unmoored from this work. We should make this code however we need it to be. `assoc_rdfgen.py` is under a totally different paradigm.

- `validate.py` should be the parts that are aware of metadata. But code in rdfgen should know the least about the context. Validate understands the context and then executes into rdfgen correctly, and rdfgen does its job - but shouldn't be very aware. Look to the separation between `produce` and the parsers: `produce` knows what the outside world wants, is aware of metadata, etc, but then once we are parsing we have setup the execution environment the way we want and it's on it's way.

dougli1sqrd:
https://github.com/geneontology/go-ontology/issues/19265

PR #480, opened 2020-09-29
Adding gocamgen emitter code

dustine32:
Current state of [gocamgen](https://github.com/geneontology/gocamgen) in `ontobio`. These changes shouldn't affect anything currently in the GO pipeline. This is just adding the files for a new N-Quads/TTL emitter along with the `gpad2gocams` command in `bin/validate.py`, which relates to #432.

Also somewhat related to https://github.com/geneontology/gocamgen/issues/70.


dougli1sqrd:
@dustine32 you're satisfied with everything here? I don't think this is directly connected to anything at the moment, so in the end this is low risk, yeah?

dustine32:
@dougli1sqrd Right, should be low risk. The only things to watch would be the added `validate.py` command `gpad2gocams`, new/changed relations, and the new `PyShEx` dependency.

I'd be OK if you merged this.

Issue #475, opened 2020-09-23
GPAD parser should set the relation in GoAssociation using the qualifier field


dougli1sqrd:
Currently GPAD 1.1 parser does not set the `relation` field. This should be grabbed from the `qualifier` field.

sierra-moxon:
fixed via PR above.

PR #481, opened 2020-09-30
gpad now correctly sets the relation when parsing

dougli1sqrd:
For https://github.com/biolink/ontobio/issues/475


kltm:
Looks fine to me. What was the thing that was borken?

dougli1sqrd:
oh just the gpad parser wasn't setting the `relation`  field in the GoAssociation, only the qualifier.

PR #484, opened 2020-10-20
Go site 1453 gpad 2 parser

dougli1sqrd:


dougli1sqrd:
For partial fulfillment of https://github.com/geneontology/go-site/issues/1453

kltm (commented):
Would it be possible to also get a non-MGI test set, or maybe that's there but not in the diff?

kltm:
I took a look and it seems to be about what I might expect. Is there an emitter associated with this?

dougli1sqrd:
Yeah, I think in master we have an emitter currently, but again, nothing is "hooked up" to the outside world. But this lets us make a command line tool or modify existing tools.

Can you clarify more on your "non MGI test set" comment? Included in this PR are unit tests of small scale functionality, but probably modifying/creating command line parsing tools that can do gpad 2 in and out maybe in a small pipeline might be a better test?

dustine32 (approved):

PR #485, opened 2020-10-22
Association collections

dougli1sqrd:
This is the bit that will help thread through arbitrary Annotation data along with Bio Entity data. Drawing when the three of us talked about this: https://docs.google.com/drawings/d/1QSurx_SG50LXUpVRUYZVWNkSF8szGrGh5DpBtXRzKGo/edit

### DIFF
--- a/bin/ontobio-parse-assocs.py
+++ b/bin/ontobio-parse-assocs.py
@@ -72,6 +72,8 @@ def main():
                         help='Increase output verbosity')
     parser.add_argument("--allow_paint", required=False, action="store_const", const=True,
                         help="Allow IBAs in parser")
+    parser.add_argument("-g", "--gpi", type=str, required=False, default=None,
+                        help="GPI file")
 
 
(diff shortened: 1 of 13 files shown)


kltm:
Excitement!

dustine32:
@dougli1sqrd Oh right, I suppose the same code change could apply to `master` right now. Should I make a separate branch and PR to `master` for this?

dougli1sqrd:
Yeah I imagine we should. This could show up in the current world now, so let's separate it out.

dougli1sqrd:
Okay, this is a real PR now!

kltm:
@dougli1sqrd What does the "risk profile" look like to you? Should this go in before the next release attempt?

dougli1sqrd:
Low risk, but we should probably go ahead and wait for the release to make an ontobio release and bump go-site. I think we'd be safe to merge to master though.

kltm:
What he sez.

PR #487, opened 2020-10-28
curies should not parse with spaces in them, tests for new validation

dougli1sqrd:


dougli1sqrd:
Just fyi, this is different than the example Kimberly showed on the GO-CAM spec call on Oct 27.

kltm:
Looks like fix plus test. :+1:

Issue #489, opened 2020-11-30
Taxon Curie objects are being leaked into the GPAD output


dustine32:
Noticed this in `goa_human.gpad` from [2020-11-17 release](http://release.geneontology.org/2020-11-17/annotations/):
```
UniProtKB       A0A087WT01      involved_in     GO:0050830      PMID:12682246   ECO:0000270             [Curie(namespace='NCBITaxon', identity='1280')] 20200113        UniProt
```
The `interacting_taxon` column value is the Curie object (`[Curie(namespace='NCBITaxon', identity='1280')]`) instead of either `NCBITaxon:1280` or `taxon:1280`. Fortunately, this same line in the GAF looks to be rendered correctly (`taxon:9606|taxon:1280`):
```
UniProtKB       A0A087WT01      TRAV27          GO:0050830      PMID:12682246   IEP             P       T cell receptor alpha variable 27       TRAV27  protein taxon:9606|taxon:1280   20200113        UniProt
```
Related to #477. I think it's due to [this line](https://github.com/biolink/ontobio/blob/8dfc25e887f1b45743215a10ad0b6251b2bbe52b/ontobio/model/association.py#L345) string-ifying a list of `Curie`'s rather than a single `Curie`.

@pgaudet @kltm @cmungall This will need to be fixed and tested before the next release.

dougli1sqrd:
Oh interesting. Yeah, somewhere we must be placing a list of one item into a `GoAssociation.interacting_taxon` rather than a single item or nothing (`None`).

dougli1sqrd:
Downloading the release `goa_human-src.gaf`, I tried to duplicate the bug:
```
$ ontobio-parse-assocs.py -r go-ontology.json --report-md rep.md -f goa_human-src.gaf -o goa_human.gpad -v convert -t gpad
$ cat goa_human.gpad | grep Curie  | wc -l
0
```
And this is using master and v2.1.0 (which is the current ontobio release)

Whereas taking snapshot or release goa_human.gpad:
```
cat goa_human.snap.gpad | grep Curie | wc -l
526
```

dougli1sqrd:
So, tracing a messed up line in snapshot, locally I see:
```
$ cat target/groups/goa/goa_human.gpad | grep 'PMID:16888103' | grep '33892'
[...]

PR #490, opened 2020-12-03
gpad parser now takes taxon element out of the list instead of passing on the entire list

dougli1sqrd:
Fixes issue #489

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -54,7 +54,7 @@ setuptools.setup(
         'marshmallow==3.0.0b11',
         'scipy',
         'pandas==1.0.5',
-        'click==7.0',
+        'click',
         'yamldown',
         'dataclasses',
         'diskcache>=4.0.0',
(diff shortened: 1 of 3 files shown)


kltm:
Looks good to me...but I'm new here. Will wait on tests.

setup.py
```diff
@@ -54,7 +54,7 @@
         'marshmallow==3.0.0b11',
         'scipy',
         'pandas==1.0.5',
```
deepakunni3, line 56: Can we also remove the version lock for Pandas? This has been a problem before when other tools use Ontobio as a dependency.

deepakunni3: And the lock on networkx too :)

deepakunni3 (changes requested):

setup.py
dougli1sqrd: oh sure

deepakunni3 (approved):

Issue #494, opened 2020-12-10
Add a version output option for ontobio-parse-assocs.py convert


dougli1sqrd:
ontobio-parse-assocs.py `convert` command needs a version option to select what version output for a file type.

PR #495, opened 2020-12-10
adding version option to convert subcommand in ontobio-parse-assocs

dougli1sqrd:
For https://github.com/biolink/ontobio/issues/494

### DIFF
--- a/bin/ontobio-parse-assocs.py
+++ b/bin/ontobio-parse-assocs.py
@@ -89,7 +89,7 @@ def main():
     parser_n.set_defaults(function=convert_assocs)
     parser_n.add_argument('-t', '--to', type=str, required=True, choices=["GAF", "GPAD"],
                           help='Format to convert to')
-    parser_n.add_argument("-v", "--version", type=str, required=False,
+    parser_n.add_argument("-n", "--format-version", type=str, required=False, default=None,
                           help="Version for the file format. GAF default is 2.1, GPAD default is 1.2")
 
     parser_n = subparsers.add_parser('map2slim', help='Map to a subset/slim')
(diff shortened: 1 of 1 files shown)


kltm:
@dougli1sqrd 
Shouldn't there be a choice of GAF and GAF 2.2? And I guess GPAD/GPI is always assumed to be "2.0", but adding versions the right way now might make it easier when the inevitable increment comes. 
Looks good to me, assuming the underlying system can deal with it.

dougli1sqrd:
For invalid versions, the default for each writer/parser is used if a weird value.

kltm:
@dougli1sqrd Right, but specifically for our use case, don't we need to be able to actively toggle between GAF 2.1 and GAF 2.2? What `-t` flag is used?

dougli1sqrd:
Yeah, so if you do `ontobio-parse-assocs.py -F gaf -f mgi.gaf -o mgi.gpad convert -t gpad -v 2.0` you'll take your mgi.gaf gaf and convert it into a gpad 2.0. If you supply something wacky for `-v` it'll just use `1.2`, since that's the default for GPAD.

And if you want GAF whichever, `ontobio-parse-assocs.py -F gaf -f mgi.gaf -o mgi.gpad convert -t gaf -v 2.1` to conver to GAF 2.1, etc.

kltm:
@dougli1sqrd Aaaahhhh--I didn't realize that format and version were separate. Nevermind!

kltm:
(I would note though, that `-v` and `--version` are usually reserved for software introspection; e.g. https://www.gnu.org/prep/standards/html_node/Command_002dLine-Interfaces.html)

dougli1sqrd:
I did think about that, but I couldn't think of a better letter ha. I can make something up tho that's maybe better than v

kltm:
`r`evision, version `n`umber?
What I was assuming, and why it was easy to not see what you were doing, is that I was expecting something like:
`gaf`, `gaf2.1`, `gaf2.2`, or the like.
I'm thinking of things like pandoc:
```bash
sjcarbon@moiraine:~/local/src/git/ontobio[gocamgen]$:) pandoc --list-output-formats
...
docbook
docbook5
...
epub
epub3
fb2
haddock
html
html5
...
markdown
markdown_github
markdown_mmd
markdown_phpextra
markdown_strict
...
```

kltm:
Of course, if this is what we've been doing, we can continue for now--we don't want to derail on what we already have.

Issue #496, opened 2020-12-10
ontobio validate.py should grabs the "compression" value from the dataset metadata without guards


dougli1sqrd:
Line 138 of validate.py we have `dataset_metadata["compression"] == "gzip"` when grabbing compression info for a dataset source.

If the `compression` field is not present -- which is valid by the schema -- should just act as empty. Additionally we should ensure that an empty value for `compression` correctly acts as "empty" or false when being investigated.

PR #497, opened 2020-12-10
protect against a missing key in dataset yaml

dougli1sqrd:
for https://github.com/biolink/ontobio/issues/496


kltm:
@dougli1sqrd Makes sense to me. Merge at will.

Issue #499, opened 2020-12-11
Gaf 2.2 writing enabled in ontobio-parse-assocs reveals lack of test and bug in gaf 2.2 writing


dougli1sqrd:
Now that ontobio-parse-assocs can specify file format version numbers, a very embarrassing bug due to a typo is revealed in the gaf 2.2 writing code.

Additionally, looking for a test that would missed this error, I found there there *weren't* any tests for writing gaf 2.2. It seems as if none were written, to my own embarrassment.

This should be fixed immediately, as the current master ontobio-parse-assocs conversion to gaf 2.2 is currently broken with this bug.

PR #500, opened 2020-12-11
bug fixes for writing gaf 2.2

dougli1sqrd:
For https://github.com/biolink/ontobio/issues/499


kltm:
@dougli1sqrd I think everything here makes sense to me--merge when you're ready.

In general, I'd go ahead and error out on bad command input (e.g. GAF 3.1) rather than falling back to
a sensible default. As this is often pipeline software, rather than interactive, it's best not to let things that
could be seen as "data loss" leak in for QC later on to discover (or not!).

dougli1sqrd:
That makes sense. This command and tool isn't used anywhere in the pipeline, and I don't think it ever will be. This is more of a human run tool these days.

Issue #493, opened 2020-12-08
Support YYYY-MM-DD format in test for GORULE:0000029


dustine32:
When parsing this GPAD 2.0 line:
```
MGI:MGI:1914305         RO:0002264      GO:0001933      MGI:MGI:5514426|PMID:23986251   ECO:0000314                     2014-05-20      MGI             creation-date=2014-05-20|modification-date=2014-05-20|contributor-id=http://orcid.org/0000-0003-3394-9805
```
I get:
```
Traceback (most recent call last):
  File "gpad2gaf.py", line 22, in <module>
    assocs: List[GoAssociation] = gpad_parser.parse(args.gpad_path, skipheader=True)
  File "/Users/ebertdu/go/ontobio/ontobio/io/assocparser.py", line 464, in parse
    a = list(associations)
  File "/Users/ebertdu/go/ontobio/ontobio/io/assocparser.py", line 477, in association_generator
    parsed_result = self.parse_line(line)
  File "/Users/ebertdu/go/ontobio/ontobio/io/gpadparser.py", line 150, in parse_line
    go_rule_results = qc.test_go_rules(assoc, self.config)
  File "/Users/ebertdu/go/ontobio/ontobio/io/qc.py", line 705, in test_go_rules
    result = rule.value.run_test(active_annotation, config, group=group)
  File "/Users/ebertdu/go/ontobio/ontobio/io/qc.py", line 87, in run_test
    result = self._run_if_context(annotation, config, group=group)
  File "/Users/ebertdu/go/ontobio/ontobio/io/qc.py", line 82, in _run_if_context
    result = self.test(annotation, config, group=group)
  File "/Users/ebertdu/go/ontobio/ontobio/io/qc.py", line 390, in test
    int(date[6:8]),
ValueError: invalid literal for int() with base 10: '5-'
```
It's choking on the `2014-05-20` date field due to the coded indices here:
https://github.com/biolink/ontobio/blob/9ac3a06bc67672e2576ced7eb8c0c232668451c6/ontobio/io/qc.py#L388-L391

So we'll need to update this to allow date format in [GPAD 2.0 spec](https://github.com/geneontology/go-annotation/blob/master/specs/gpad-gpi-2-0.md#gpad-columns)

@dougli1sqrd I can PR a change and test.

dougli1sqrd:
oh thanks, good catch!

sierra-moxon:
fixed via PR above.

PR #502, opened 2020-12-16
Pass normalized date to GoAssoc for #493

dustine32:
Uses the `date` variable, normalized with date format (YYYYMMDD), for internal GORULE tests rather than just passing through raw column value `gpad_line[8]`

### DIFF
--- a/ontobio/io/assocparser.py
+++ b/ontobio/io/assocparser.py
@@ -830,25 +830,70 @@ class AssocParser(object):
 
 
 
-def _normalize_gaf_date(date, report, taxon, line):
-    if date is None or date == "":
-        report.warning(line, Report.INVALID_DATE, date, "GORULE:0000001: empty",
-            taxon=taxon, rule=1)
-        return date
+def parse_date(date: str, report: Report, line: List) -> Optional[association.Date]:
+    if date == "":
+        report.error(line, Report.INVALID_DATE, "\'\'", "GORULE:0000001: empty", rule=1)
+        return None
 
-    # We check int(date)
+    d = None
     if len(date) == 8 and date.isdigit():
-        d = association.Date(date[0:4], date[4:6], date[6:8], "")
+        # For GAF date, should be YYYYMMDD all as digits and
+        # a well formed date string here will be exactly 8 characters long
+        d = association.Date(year=date[0:4], month=date[4:6], day=date[6:8], time="")
     else:
-        report.warning(line, Report.INVALID_DATE, date, "GORULE:0000001: Date field must be YYYYMMDD, got: {}".format(date),
-            taxon=taxon, rule=1)
+        report.warning(line, report.INVALID_DATE, date, "GORULE:0000001: Date field must be YYYYMMDD, got: {}".format(date), rule=1)
+        parsed = None
         try:
-            d = dateutil.parser.parse(date)
-            d = association.Date(str(d.year), str(d.month), str(d.day), "")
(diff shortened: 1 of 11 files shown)


dustine32:
Actually, this doesn't fix for GPAD 2.0 since it's not changed in `from_2_0`. Also, normalizing this format (YYYY-MM-DD) will transform to incorrect format (YYYYMMDD) according to 2.0 spec.

More work needed to fix.

dustine32:
I believe https://github.com/biolink/ontobio/pull/502/commits/62327aed7835af72775ac0a6809037ebca7f22df should fix #493. I also fleshed out the test to trigger the GPAD 2.0 parse code.

Even though it doesn't affect this issue, I'm leaving the `gpadparser.py` change in since I assume it was intended to set the normalized date in the `GoAssociation` object. @dougli1sqrd What do you think?

dougli1sqrd:
So the best way to solve this will be to figure out what the "truest" representation of the Date should be, and always parse from an annotation file into the real Date, in the `GoAssociation`. The gorule should be able to trust and rely completely on the correctness of the date format.

We could use the builtin datetime formats for this, but I fear those could be slow. In the past parsing dates was slow, and why we deferred to this semi-hard-coded string-as-array method.

Maybe for speed and expressiveness we could make a `collections.namedtuple` with fields like `day, month, year`. Although we may have to support some sub spec of ISO 8601. Maybe there's some optimization we can make that is easy to convert from a local domain type into the builtin datetime side? They definitely support ISO 8601, so probably that's where we ultimately want to go.

@kltm Where was the ticket (was there one?) describing the ISO 8601 date possible requirements for the future?

dougli1sqrd:
Talking with Seth yesterday, I'll record a summary here of what we talked about.

When parsing the date from an incoming annotation file, we need to validate it, and perhaps upgrade it to the right format (`YYYY-MM-DD`). But during the imports, an optional `THH:MM` may be appended. Essentially this should be passed through for use by making GO-CAMs.

The key requirement here is that *after* parse, the date must be in the standardized `YYYY-MM-DD[THH:MM]` format. For parsing purposes though, we only care about the date portion.

So, since each annotation format essentially has its own date format (or rather, since we have more than one now, we can't assume the same format anymore), we should have a each `to_association` associated with a format do the correct conversion, with the assumption first that it's correct, and then falling into the builtin `datetime` parsing (like we do currently with gaf dates) if the simple way doesn't work.

I see us as having two choices as to what to parse into 
1) retain the raw string and just pass it through
2) parse into a lightweight type that gives us *some* type documentation.

I think using **2** would be pretty straightforward, and would enhance readability and express intent. `namedtuple` is a good choice here since it really is just a tuple, which are  fast in python and isn't the same as creating a full class. We have had issues in the past with slow parsing of dates in the parser before. `namedtuple` is really just a thin wrapper around a regular `tuple` so has builtin speed this way.

We could make a `namedtuple` AssociationDate with fields `year`, `month`, `day`, `time`. In the ideal cases where the incoming date strings are well formed, we can just use array indexing to find these fields (essentially like already happens). Then falling through to the builtin datetime parsers, to grab the right fields. The `time` element should just be sliced off the end, like `date[10:0]` which will grab the 10 index and after.

dougli1sqrd:
I think this is good with my update. Unless there's anything else you want to add @dustine32, I think we should merge this.

dustine32:
@dougli1sqrd Agreed! I just now pulled, tested and everything seems to work fine. Thanks again for fixing this!

Issue #504, opened 2021-01-05
Ontology.merge() should copy logical definitions and property chain axioms


dustine32:
The `merge` function on `Ontology` https://github.com/biolink/ontobio/blob/61878238d0d6eb4dccb4b48a285f7b9ae5eac5f4/ontobio/ontol.py#L136 ...should transfer over all logical definitions and property chain axioms contained in supplied `ontologies`.

My use case here is merging RO into GO for use in `gocamgen`:
```
ontology_graph = OntologyFactory().create("go.json")
ontology_graph.merge(OntologyFactory().create("ro.json"))
pcas = ontology_graph.get_property_chain_axioms("RO:0002264") # acts upstream of or within
causally_upstream_relation = pcas[0].chain_predicate_ids[1] # causally upstream of or within
```
@dougli1sqrd I discovered this was missing when attempting to create our generic `ontology` object. I already have a fix with some tests to PR.

PR #505, opened 2021-01-05
Ontology.merge copies logical defs and property chain axioms

dustine32:
For issue #504.

### DIFF
--- a/ontobio/ontol.py
+++ b/ontobio/ontol.py
@@ -151,6 +151,12 @@ class Ontology():
             if ont.xref_graph is not None:
                 for (o,s,m) in ont.xref_graph.edges(data=True):
                     self.xref_graph.add_edge(o,s,**m)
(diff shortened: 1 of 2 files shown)


dustine32:
@cmungall Can you check out this PR to see if it makes sense?

It's code to pull all logical definitions and property chain axioms from `ontology_b` when doing `ontology_a.merge(ontology_b)`. Or for my use case:
```
ontology = go_ontology.merge(ro_ontology)
```
This single "super"-ontology is then used by gocamgen in places like:
https://github.com/biolink/ontobio/blob/af847c277cc33094972993e1aaa58ce7f9639ea9/ontobio/rdfgen/gocamgen/gocamgen.py#L753

dougli1sqrd:
Tagging @kltm

cmungall (approved):

dougli1sqrd (approved):

Issue #482, opened 2020-10-13
Ontobio should not pin dependencies to a particular version


deepakunni3:
Currently the latest release of Ontobio has the following dependencies pinned to a fixed version:
```
marshmallow==3.0.0b11
pysolr==3.6.0
networkx==2.2
matplotlib==2.0.0
SPARQLWrapper==1.8.0
scipy==1.2.0
plotly==2.0.7
```

This creates problems when Ontobio is being used as a dependency for downstream applications.
Is it possible to have all these be `>=`?
Or are there unexpected side effects that disrupts Ontobio's functionality?

dhimmel:
Looking through `install_requires`, I noticed the same thing (several pinned dependencies):

https://github.com/biolink/ontobio/blob/61878238d0d6eb4dccb4b48a285f7b9ae5eac5f4/setup.py#L46-L62

This makes ontobio likely to be incompatible with many user environments. For example, networkx is currently at version 2.5.

Solutions might be:

- pin install_requires only to major versions or nothing at all.
- run scheduled CI jobs that will detect if dependency updates break things.

matentzn:
Is there anyone that can look into this? It really makes ontobio incompatible with many other libraries.

matentzn:
(For me its networkx)

deepakunni3:
@matentzn https://github.com/biolink/ontobio/pull/585 should fix the above issue

PR #508, opened 2021-01-09
Fix dependencies in requirements.txt and setup.py

### LINKED ISSUE #506: requirements.txt requiring networkx 2.2
Hi Deepak,

Just an FYI, the `requirements.txt` file for ontobio (in the master branch) appears to be hard-coded to use networkx 2.2, but that version of networkx is (I think) not compatible with python3.9. It looks like you recently merged in some code changes to update ontobio's usage of networkx; am wondering if you can update the requirements.txt accordingly?  Am assuming that it should not be `==2.2` ?

Kindly,
Steve & Team Expander Agent

deepakunni3:
This PR addresses the following,
- Fixes dependency versions
- Removes duplicated declaration of dependencies in `setup.py` and read from `requirements.txt`

Fixes #506 #482


cmungall:
Looks good. We should have done the setup.py thing ages ago. Isn't there a version of networkx we need ro exclude?

deepakunni3:
Ontobio now would work with all versions of networkx that start with 2.x since there is no use of deprecated/removed methods.

dougli1sqrd:
Is it normal to to read from requirements like this? Maybe we should unify these by only using setup.py as it's the standard way of installing ontobio anyway. Do we even want people to run requirements.txt by itself? Just feels a little magical.

cmungall:
I think this PR is a big improvement, both in unpining specific versions, and reducing redundancy

Is there any reason not to merge?

maybe bring up your proposal in a separate ticket @dougli1sqrd ? Whatever solution we choose I want it to be the same across all our python projects

dougli1sqrd:
I don't know that I have a specific proposal. I think from our convo a few weeks ago about this using setup to read the requirements so that installing in both methods to support people who want to do both seems fine. I was just feeling cautious since I hadn't seen that before in other python projects generally.

@deepakunni3 I think if we fix the conflicts we can just merge.

deepakunni3:
Thanks @dougli1sqrd !

jLebioda:
Hi @deepakunni3 !
It looks like you kept `marshmallow` pinned to `3.0.0b11`, and that creates a problem for Python 3.10:

```
  File "/usr/local/lib/python3.10/site-packages/marshmallow/schema.py", line 5, in <module>
    from collections import defaultdict, Mapping, OrderedDict
ImportError: cannot import name 'Mapping' from 'collections' (/usr/local/lib/python3.10/collections/__init__.py)
```

This is related to a bug that was already fixed in Marshmallow - https://github.com/marshmallow-code/marshmallow/issues/1174

Could you please bump the _marshmallow_ library in requirements, please?

jLebioda:
> Hi @deepakunni3 ! It looks like you kept `marshmallow` pinned to `3.0.0b11`, and that creates a problem for Python 3.10:
> 
> ```
>   File "/usr/local/lib/python3.10/site-packages/marshmallow/schema.py", line 5, in <module>
>     from collections import defaultdict, Mapping, OrderedDict
> ImportError: cannot import name 'Mapping' from 'collections' (/usr/local/lib/python3.10/collections/__init__.py)
> ```
> 
> This is related to a bug that was already fixed in Marshmallow - [marshmallow-code/marshmallow#1174](https://github.com/marshmallow-code/marshmallow/issues/1174)
> 
> Could you please bump the _marshmallow_ library in requirements, please?

Ahh, sorry! It looks like it's already been done; I've tested accidentally the previous version (2.7. instead of 2.8). All is fine; my bad!

PR #509, opened 2021-01-27
updating gpi parsing to allow for version 2.0. Tests for changes

dougli1sqrd:
This change allows parsers to parse GPI 2.0 as well when processing GPAD/GAF. Currently this is just for *reading*, and does not write GPI 2.0 yet.

The changes are tested, I plan on doing some more tests replicating the pipeline to make sure everything is good.


kltm:
A quick look seems to reveal nothing not expected, with tests included. Seems good to me.
Would like input from @dustine32

dougli1sqrd:
@dustine32 and I looked through some files we ran using this PR to detect any anomolies in gpad 2.0 parsing and output, as well as the pipeline `validate.py` script and are satisfied that this is working properly.

dustine32 (approved):
@dougli1sqrd This is real good! I don't see anything obviously wrong here.

PR #515, opened 2021-02-03
threading gaf output version number through validate and small bugfix in data output

dougli1sqrd:
For https://github.com/geneontology/pipeline/issues/211


dustine32 (approved):
@dougli1sqrd This all looks great!

PR #517, opened 2021-02-09
Go-site issue 1558 gorule 59

dougli1sqrd:
gorule-0000059 to upgrade gaf 2.1 qualifiers into better relations based on GO Term.

https://github.com/geneontology/go-site/issues/1558

### DIFF
--- a/ontobio/io/gafparser.py
+++ b/ontobio/io/gafparser.py
@@ -37,8 +37,10 @@ gaf_line_validators = {
     "taxon": assocparser.TaxonValidator()
 }
 
-def compute_cell_component_subtree_closures(ontology: Ontology) -> Dict[association.Curie, Set[association.Curie]]:
-    ontology.descendants()
+def protein_complex_sublcass_closure(ontology: Ontology) -> Set[str]:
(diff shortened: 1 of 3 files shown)


ontobio/io/gafparser.py
```diff
@@ -144,13 +156,16 @@ def parse_line(self, line):
                     else:
                         logger.info("Detected GAF version {}, so using 2.1".format(version))
                         self.version = self.default_version
+                        # Compute the cell component subclass closure
```
dustine32, line 159: @dougli1sqrd Does this here mean the closure will be recalculated on every call of `parse_line` for a 2.1 GAF?

dougli1sqrd: No, this only computes the closure if we are in the headers and once we have detected that we are in gaf 2.1. So this should happen only once per gaf 2.1 file.

dustine32: Ah, right. I just realized that after staring harder. All is good, thanks!

dustine32 (approved):

kltm:
@dougli1sqrd Happy to see merged.

Issue #513, opened 2021-02-02
Annotation properties should allow multiple values of the same key


dustine32:
For this [GPAD 2.0](https://github.com/geneontology/go-annotation/blob/master/specs/gpad-gpi-2-0.md) line:
```
MGI:MGI:97591           RO:0001025      GO:0005929      MGI:MGI:3654822|PMID:16687649   ECO:0000314        2013-02-06      MGI     BFO:0000050(CL:0000019) creation-date=2008-09-25|modification-date=2013-02-06|contributor-id=http://orcid.org/0000-0003-2689-5511|contributor-id=http://orcid.org/0000-0002-9796-7693
```
Where annotation properties column (col 12) contains two `contributor-id` key-value pairs:
```
contributor-id=http://orcid.org/0000-0003-2689-5511|contributor-id=http://orcid.org/0000-0002-9796-7693
```
Parsers, writers, and the `GoAssociation` model should support both of these `contributor-id` values. But currently we just construct a dictionary by iterating through the properties, overwriting any previously encountered keys (e.g. `contributor-id`, `comment`):
https://github.com/biolink/ontobio/blob/de97ea7a59ff3d7d3305e941742b5790b264c25f/ontobio/io/gpadparser.py#L462
This is kind of like:
```
properties["contributor-id"] = "http://orcid.org/0000-0003-2689-5511"
properties["contributor-id"] = "http://orcid.org/0000-0002-9796-7693"
```
As @dougli1sqrd suggested, we could switch this dictionary to just be a list of key-value tuples, similar to a few lines before this:
https://github.com/biolink/ontobio/blob/de97ea7a59ff3d7d3305e941742b5790b264c25f/ontobio/io/gpadparser.py#L446
But now like:
```python
properties_list = [tuple(prop.split("=", maxsplit=1)) for prop in gpad_line[11].split("|") if prop]
```
And this would retain the multiple `contributer-id` data:
```
[
   ('creation-date', '2008-09-25'), 
   ('modification-date', '2013-02-06'), 
   ('contributor-id', 'http://orcid.org/0000-0003-2689-5511'), 
   ('contributor-id', 'http://orcid.org/0000-0002-9796-7693')
]
```
Of course, any code that uses annotation properties (e.g. GPAD writers, gocamgen) will have to adjust to the new DS.

dougli1sqrd:
This is great!

[...]

PR #524, opened 2021-02-16
Issue #513 - change annotation props from dict to list of 2-tuples

dustine32:
@dougli1sqrd Here's my first go! Let me know if I should add more tests. Do you think we should make some sort of class for this new `properties: List[Tuple[str, str]]` data structure?

### DIFF
--- a/tests/test_gpad_parser.py
+++ b/tests/test_gpad_parser.py
@@ -59,6 +59,11 @@ def test_parse_interacting_taxon():
     result = to_association(list(vals), report=report, version="1.2")
     assert result.associations[0].interacting_taxon == Curie(namespace="NCBITaxon", identity="5678")
 
+def test_duplicate_key_annot_properties():
+    properties_str = "creation-date=2008-02-07|modification-date=2010-12-01|comment=v-KIND domain binding of Kndc1;MGI:1923734|contributor-id=http://orcid.org/0000-0003-2689-5511|contributor-id=http://orcid.org/0000-0003-3394-9805"
+    prop_list = assocparser.parse_annotation_properties(properties_str)
+    contributor_ids = [value for key, value in prop_list if key == "contributor-id"]
+    assert len(contributor_ids) == 2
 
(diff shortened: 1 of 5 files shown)


tests/test_gpad_parser.py
```diff
@@ -59,6 +59,11 @@ def test_parse_interacting_taxon():
     result = to_association(list(vals), report=report, version="1.2")
     assert result.associations[0].interacting_taxon == Curie(namespace="NCBITaxon", identity="5678")
 
+def test_duplicate_key_annot_properties():
+    properties_str = "creation-date=2008-02-07|modification-date=2010-12-01|comment=v-KIND domain binding of Kndc1;MGI:1923734|contributor-id=http://orcid.org/0000-0003-2689-5511|contributor-id=http://orcid.org/0000-0003-3394-9805"
+    prop_list = assocparser.parse_annotation_properties(properties_str)
+    contributor_ids = [value for key, value in prop_list if key == "contributor-id"]
+    assert len(contributor_ids) == 2
```
dougli1sqrd, line 66: For our tests, let's try to test specific values as much as we can. Maybe we can put the contributor properties tuples into a set, and then assert that our expected contributors match what we found in the line.

tests/test_gpad_parser.py
```diff
@@ -112,3 +117,7 @@ def test_parse_2_0():
     vals[0] = "WB:WBGene00001189"
     result = to_association(list(vals), report=report, version=version)
     assert result.associations[0].subject.id == Curie("WB", "WBGene00001189")
+
+    # Test annotation property retrieval
+    contributors = result.associations[0].annotation_property_values(property_key="contributor-id")
```
dougli1sqrd, line 122: Same suggestion here as in `test_duplicate_key_annot_properties`

github is also complaining of lack of new line at the end of this file?

dougli1sqrd (changes requested):

dougli1sqrd (approved):

Issue #525, opened 2021-02-16
Report should only receive strings and not objects


dougli1sqrd:
There are a couple places during gaf/gpad parsing where objects are passed into the reports.

These instances should be cleaned up and only pass in strings. This has effects with writing the reports into JSON.

sierra-moxon:
This appears to have been fixed in the related PR - closing? :)

PR #526, opened 2021-02-16
report message calls make sure to wrap taxon objects as strings

dougli1sqrd:
For https://github.com/biolink/ontobio/issues/525


dustine32 (approved):
@dougli1sqrd Looks good to me!

PR #530, opened 2021-02-25
expand a TSV using ontology lookups

cmungall:
@deepakunni3 @kshefchek - example of a fast method of enriching a tsv - e.g. kgx - with ontologies, for loading into a denormalized solr store or similar


cmungall:
this could also be combined with a step to encode categories, see https://pbpython.com/categorical-encoding.html

PR #531, opened 2021-02-26
updating gorule 59 upconvert from gaf 2.1

dougli1sqrd:


kltm:
Root tests and updates seem to be in line with rule 59 updates.

PR #537, opened 2021-03-10
updating gorule 61 to repair in bio process

dougli1sqrd:


kltm:
Tests and code update.

PR #542, opened 2021-03-22
Gocamgen reporting changes and deleting gpad_extensions_mapper

dustine32:
This writes out the GPAD parser GO rules report in `gpad2gocams` along with adding an `errors` container for passing model translation error data (e.g. Couldn't get regulates relation from logical definition of a term) to the translation report.

Also, completely dropped `gpad_extensions_mapper.py` and refactored the remaining (though already commented out) gocamgen code that was referencing it. Removing the rigged `is_cool` conditional caused the enormous tab-shift as seen in the diff.

@kltm According to @dougli1sqrd, the code I just removed was initially causing some problems in testing of ontobio's static analysis tools. Do you have a way for us to test this is fixed now? Eric pasted the error results of some `isolate following_rules` command?


kltm:
@dustine32 For this merge, I'll defer to @dougli1sqrd .

I wouldn't worry too much about the static code testing for the moment, but you can see some of the commands and where this is coming from over here: https://github.com/berkeleybop/bbops/issues/6

PR #543, opened 2021-03-22
Emit comments from GPAD annotation properties

dustine32:
Should fix geneontology/gocamgen#87.

This will emit comments from GPAD annotation properties `comment` field to the translated GO-CAM.


kltm:
Pinging @dougli1sqrd .

Issue #544, opened 2021-03-26
Is there a way to exclude some GO rules when parsing a GAF file?


valearna:
To generate the Alliance gene descriptions, I'm loading GO obo file and the GAF files for the MODs with ontobio, but GO rule 26 is skipping IBA annotations. Is there a way to exclude a specific rule when parsing the GAFs? I'm happy to work on a PR if needed.

dustine32:
Hey @valearna! I used to have this problem too. Somewhere in the `AssocParserConfig` there is a `paint` attribute that can be set to `True` to allow IBA annotations. Oh here it is:
https://github.com/biolink/ontobio/blob/0ec3aa6fea9d4492a9873a4b9b394c4866f741b6/ontobio/io/assocparser.py#L197
Let me know if you need more info!

valearna:
Thanks @dustine32!

cmungall:
I find it surprising that filtering is the default behavior!

GO team - remember ontobio is a general purpose framework that includes functionality for things like parsing GAFs. We of course want to use this general purpose framework to do things like implement specific rules such as filtering IBA from non-paint sources when run in a specific context.

Even for our own internal GOC purposes this default behavior could come back and bite us. Let us say we rewrite the solr loadiing in python. If we load the outputs of the pipeline then if we do not remember the magic flag we will accidentally filter

@valearna - thanks for the report we will prioritize fixing the behavior!

kltm:
To clarify maybe, the general/default case should not apply the rules, the optional/GO cases should be explicitly turned on.
A fix here for us then would look like shifting the default behavior and adding the hooks to turn what we want on at need.

valearna:
Thanks again for checking my PR on agr_loader @cmungall. I've set the attribute `paint=True` and now everything works as expected. I agree that it would be much more intuitive to have the attribute set to True by default.

cmungall:
...and we should use a more meaningful flag name! `paint=True` is the height of GO jargon!

PR #545, opened 2021-04-09
Issue 544 rulesets

dougli1sqrd:
For https://github.com/biolink/ontobio/issues/544


kltm:
Looks good to me so far.
Is the idea that if you declare rules, but don't specify, they all run, or does one need to be explicit?
I'll leave it to @dustine32 to do final review and merge.

dustine32 (approved):
Yeah this looks pretty cool.

dougli1sqrd:
Yeah, in ontobio-parse-assocs if no rules are specified, then none are run. You can pass in the string "all" to set all the rules, otherwise, it just uses whatever set you pass in via repeated `-l <N>`

kltm:
Okay--sounds good. Looks good to me.

PR #550, opened 2021-04-27
solr search query add support for filtering based on equivalent curie prefix, leaf nodes

kshefchek:
Adds support for filtering in/out a prefix in a equivalent curie as well as filtering for only leaf nodes. See https://github.com/biolink/biolink-api/issues/368

### DIFF
--- a/ontobio/golr/golr_query.py
+++ b/ontobio/golr/golr_query.py
@@ -347,6 +347,8 @@ class GolrSearchQuery(GolrAbstractQuery):
                  taxon=None,
                  min_match=None,
                  minimal_tokenizer=False,
+                 include_eqs=False,
+                 exclude_groups=False,
                  user_agent=None):
         self.term = term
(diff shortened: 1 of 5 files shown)


kshefchek:
@iimpulse could you review?

iimpulse (approved):

iimpulse:
out of curiosity do we have any tests?

kshefchek:
there are no tests, a little challenging to mock the solr request, but we could mock the solr response to make sure the _process_autocomplete_results fx is working as expected

iimpulse:
I was thinking more on the golr query that it builds that fq properly or that complex equivalent stuff correctly, less so on a functional or integration

kshefchek:
ah yes, that's a good idea, I'll add it before merging

kshefchek:
@iimpulse tests added with add608c

PR #552, opened 2021-05-06
Emit providedBy to all individuals; for geneontology/gocamgen#89

dustine32:
Fixes an issue where "Assigned By" isn't appearing for MOD import annotations in ART or Noctua Form.


kltm:
Single to list, looks good to me. In general, are there tests for these kinds of things?

kltm:
@dustine32 Cannot link through to issue through title geneontology/gocamgen#89

kltm (approved):

dustine32:
@kltm Thanks for fixing the link, I forgot it won't do it automatically through the title.

For tests, I used to have them for things like this on the old gocamgen repo, where I made the model in `rdflib.graph` and then SPARQL'ed out what I was looking for. But the startup cost for generating a model has become >10 sec. Maybe I can strip down the process a bit.

kltm:
Something to pick at anyways at some point. Or something.
I'm not sure if you actually want feedback from both of us; feel free to merge.

dougli1sqrd (approved):

PR #553, opened 2021-05-06
Pulling gene connection from GPI Encoded By field; for geneontology/g…

dustine32:
…ocamgen#83

Fixes issue where MOD import annotations weren't being added to the correct gene model.


kltm:
@dustine32 issue linking broken across title.
geneontology/gocamgen#83

kltm (approved):

Issue #564, opened 2021-05-24
Default rule_set in validate.py produce is parsed as list ['a', 'l', 'l']


dustine32:
This new line in `validate.py -v produce` has the default meant to trigger ALL GORULES to run:
https://github.com/biolink/ontobio/blob/aa1346c9ca92bb5a647d1c94008fb3d85c242985/bin/validate.py#L482
However, the `RuleSet.ALL` `str` is split by the `click` [multiple option handling](https://click.palletsprojects.com/en/7.x/options/#multiple-options) into `['a', 'l', 'l']` instead of the intended `['all']`. Because of this, no GORULES run by default. This default value should instead be in a list:
```
default=[assocparser.RuleSet.ALL]
```

dustine32:
Also needed to handle default `rule_set` in function since list `[assocparser.RuleSet.ALL]` is converted to tuple `(assocparser.RuleSet.ALL,)` by `click`.

dustine32:
Closing as this appears to be fixed, given that the GO rules have been running in the last few snapshot releases.

PR #565, opened 2021-05-24
Fix default --rule-set handling for #564

dustine32:
For #564


kltm:
Seems like a straightforward click tweak.

kltm (approved):

Issue #566, opened 2021-05-25
Add ns2:evidence-with property to models that have been converted to 'has_input' edges.


sierra-moxon:
From David H.  (also happening in the ZFIN import)

When an annotation is made in the conventional way to a protein binding term, the binding partner is put in the 'inferred from/with' field of the GPAD file. This is converted in an input of the binding in a GO-CAM model. We need to convert it back to the 'inferred from/with' field so the annotations match the old convention. See examples below:
https://urldefense.com/v3/__http://noctua-dev.berkeleybop.org/editor/graph/gomodel:MGI_MGI_2441920__;!!C5qS4YX3!XPuTk7YPGZkjcq1wIlpH4MMAk2qLoihBBlXCSs9ilwJyh2JrKdPq33UPhyQBufko$

input GPAD2.0
MGI:MGI:2441920         RO:0002327      GO:0005515      MGI:MGI:5574544|PMID:24843136   ECO:0000353     PR:P70338|PR:O70237             2015-04-24      MGI             creation-date=2015-04-24|modification-date=2015-04-24|contributor-id=https://urldefense.com/v3/__http://orcid.org/0000-0001-7476-6306__;!!C5qS4YX3!XPuTk7YPGZkjcq1wIlpH4MMAk2qLoihBBlXCSs9ilwJyh2JrKdPq33UPhxXs-hyO$                                    
MGI:MGI:2441920         RO:0002327      GO:0019899      MGI:MGI:5574544|PMID:24843136   ECO:0000353     PR:Q6ZQ88               2015-04-24      MGI             creation-date=2015-04-24|modification-date=2015-04-24|contributor-id=https://urldefense.com/v3/__http://orcid.org/0000-0001-7476-6306__;!!C5qS4YX3!XPuTk7YPGZkjcq1wIlpH4MMAk2qLoihBBlXCSs9ilwJyh2JrKdPq33UPhxXs-hyO$                                    

Noctua output GPAD1.1 (Note that the pipe-delimited with field is also correctly split into two annotations)
[...]

PR #567, opened 2021-05-25
#566 add with/from back to has_input models when the term is a binding term

sierra-moxon:


ontobio/rdfgen/gocamgen/collapsed_assoc.py
```diff
@@ -66,7 +66,10 @@ def collapse_annotations(self, associations: List[GoAssociation]):
             for wf in separated_with_froms:
                 ca = self.find_or_create_collapsed_association(a, wf)
                 association_line = CollapsedAssociationLine(a, wf.line)
+                # adding binding annotation with/from to the evidence line as well
+                association_line2 = CollapsedAssociationLine(a, ca.with_from())
                 ca.lines.append(association_line)
+                ca.lines.append(association_line2)
```
dustine32, line 72: I think this is creating a duplicate evidence individual, based on the test model I examined:
image
Instead of making the extra `CollapsedAssociationLine` here, I think we can try tweaking the `add_to_header` vs `add_to_line` conditional below in `get_with_froms` (see other comment).

dustine32: Grr! Dang github won't let me comment outside of changed code. I'll try this:
```
if wf_entity and wf_entity.get("taxon") == subject_entity["taxon"]:
    wf_separated.add_to_header(wf_id_str)
else:
    wf_separated.add_to_line(wf_id_str)
```
Header is basically what's getting converted to a `has_input`. Line is the "with/from" that shows up on the evidence. Without testing, I'm gonna assume, if we just want to always keep the "with/from" on the evidence, we just need to remove the `else:` so that `wf_separated.add_to_line(wf_id_str)` always runs. Like:
```
if wf_entity and wf_entity.get("taxon") == subject_entity["taxon"]:
    wf_separated.add_to_header(wf_id_str)
wf_separated.add_to_line(wf_id_str)
```
@sierra-moxon We can quick-confirm a model in my test Noctua tomorrow.

dustine32:
@sierra-moxon Looks like it worked!
image

dustine32 (approved):

Issue #568, opened 2021-05-26
Fix circular import reference between collapsed_assoc and errors


dustine32:
Running `make test` produces this error:
```
ERROR tests/test_gocamgen.py - AttributeError: module 'ontobio.rdfgen.gocamgen.collapsed_assoc' has no attribute 'CollapsedAssocia...
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
```
I traced this to `gocamgen.collapsed_assoc` importing `gocamgen.errors`, which imports `gocamgen.collapsed_assoc`, and so on...
https://github.com/biolink/ontobio/blob/f7c8d25aba9e69a830b7ac15b89185609d2ad314/ontobio/rdfgen/gocamgen/collapsed_assoc.py#L5
https://github.com/biolink/ontobio/blob/f7c8d25aba9e69a830b7ac15b89185609d2ad314/ontobio/rdfgen/gocamgen/errors.py#L1
https://github.com/biolink/ontobio/blob/f7c8d25aba9e69a830b7ac15b89185609d2ad314/ontobio/rdfgen/gocamgen/errors.py#L21

Just need to reorganize a bit.

Tagging @sierra-moxon

PR #569, opened 2021-05-26
Fix circular reference for #568; add test_gocamgen to travis

dustine32:
For #568


dustine32:
Previously erring `test_gocamgen.py` passes now.

kltm:
Looks good to me, sounds like circular reference fixed.

PR #570, opened 2021-05-27
Activate Ref Pref and add test for geneontology/gocamgen#31

dustine32:
For geneontology/gocamgen#31

### DIFF
--- a/tests/test_gocamgen.py
+++ b/tests/test_gocamgen.py
@@ -71,6 +71,15 @@ def test_ref_picker():
     result = gocamgen.ReferencePreference.pick(test_refs)
     assert result == "PMID:9834189"
 
+    test_refs = [
+        "GO_REF:0000483",
+        "doi:485930",
+        "PMID:9834189",
+        "WB_REF:WBPaper00003384",
+    ]
+    result = gocamgen.ReferencePreference.pick(test_refs)
+    assert result == "PMID:9834189"
+
     test_refs = ["ZFIN:ZDB-PUB-170709-3"]
     result = gocamgen.ReferencePreference.pick(test_refs)
     assert result == "ZFIN:ZDB-PUB-170709-3"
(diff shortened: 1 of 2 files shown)


tests/test_gocamgen.py
```diff
@@ -59,3 +59,18 @@ def test_get_with_froms():
...
     ca_set.find_or_create_collapsed_association(go_assoc, with_from=header2)
     assert len(ca_set.collapsed_associations) == 1 and ca_set.collapsed_associations[0].with_froms == ["FAKE:12345", "MGI:MGI:1915834"]
+
+
+def test_ref_picker():
+    test_refs = [
+        "GO_REF:0000483",
+        "doi:485930",
+        "WB_REF:WBPaper00003384",
+        "PMID:9834189",
```
kltm, line 69: Would it be possible to flip the PMID and WB_REF?

tests/test_gocamgen.py
```diff
@@ -59,3 +59,18 @@ def test_get_with_froms():
...
+    ]
+    result = gocamgen.ReferencePreference.pick(test_refs)
+    assert result == "PMID:9834189"
+
+    test_refs = ["ZFIN:ZDB-PUB-170709-3"]
+    result = gocamgen.ReferencePreference.pick(test_refs)
+    assert result == "ZFIN:ZDB-PUB-170709-3"
```
kltm, line 76: Question: what would happen if there were no refs? Is that a possibility?

tests/test_gocamgen.py
dustine32: Like, switch in the test to just ensure the `PMID` isn't picked because it's last?
```
        "GO_REF:0000483",
        "doi:485930",
        "PMID:9834189",
        "WB_REF:WBPaper00003384",
```

tests/test_gocamgen.py
dustine32: There would be an `IndexError` for empty list. But there should always be _something_ in references once it got to this point, otherwise it would fail elsewhere in gocamgen or get chucked out by the GPAD parser earlier.

tests/test_gocamgen.py
kltm: Yeah, sorry, I just noticed that your tests would pass if the algorithm of "pick last" and that made me feel weird. Personal problem, but I couldn't get the idea unstuck from my mind.

dustine32: Good point! I added that flip case to help make it clear.

kltm (approved):

PR #571, opened 2021-06-04
Param to set modelstate for geneontology/gocamgen#91

dustine32:
For geneontology/gocamgen#91.

To use, just need to set `--modelstate` or `-s` to `production`. Default is `development`.

### DIFF
--- a/bin/validate.py
+++ b/bin/validate.py
@@ -569,7 +569,8 @@ def produce(ctx, group, metadata_dir, gpad, ttl, target, ontology, exclude, base
 @click.option("--target", "-t", type=click.Path(), required=True)
 @click.option("--ontology", "-o", type=click.Path(exists=True), required=True, multiple=True)
 @click.option("--ttl", default=False, is_flag=True)
-def gpad2gocams(ctx, gpad_path, gpi_path, target, ontology, ttl):
+@click.option("--modelstate", "-s", default=None)
+def gpad2gocams(ctx, gpad_path, gpi_path, target, ontology, ttl, modelstate):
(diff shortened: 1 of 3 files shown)


dustine32:
BTW I tested this by loading a few test models to local Noctua and checking the modelstate. Pretty straight-forward. But for unit tests on this change, this is another case where we'd need to generate a model and then check the attribute (i.e. `modelstate`). Since model generation requires the loaded ontology, which can take 10-30 sec to load, I really need to make a stripped-down, "targeted" ontology that loads super-quick and is small enough to store it with the repo. I may make a ticket just on this problem.

kltm:
@dustine32 Looks good, but I had a question. Looking at your code, it seems that the default state is only assigned in the cam writer, with it being `None` elsewhere? Might it be an idea to have that default be `development` higher up in the code so that other future writers and subsystems would have access?

dustine32:
@kltm Yeah, I was thinking of moving up where that default is defined. Probably one level up in `GoCamModel`? I suppose `GoCamModel` is the lowest class where someone would care what/if `modelstate` is emitted to the model.

I can also consolidate how `modelstate` is passed to `GoCamBuilder` so that it's not passed through so many functions.

kltm:
@dustine32 Whatever you think is sensible, I wanted to see where your thoughts are. I think that you're right that the go-cam model would be the place where it starts mattering. I'm just hoping to cut off future spooky issues.

kltm:
Sounds goo. Feel free to merge when you want.

kltm (approved):

Issue #572, opened 2021-06-09
Change CURIE of exists_during in lookup


dustine32:
The relation `exists_during` has recently been minted an RO term ID that's now being used in the MGI import GPAD. We need to change the value in the `relations.py` lookup from http://purl.obolibrary.org/obo/GOREL_0000032 to http://purl.obolibrary.org/obo/RO_0002491.

sierra-moxon:
I see this change in relations.py, and I don't find GOREL_0000032 -- closing? :)

PR #573, opened 2021-06-09
Updated to new RO term and relaxed error handling for missing ext rel

dustine32:
For #572.

I also changed how missing extension relation errors are handled. This change now skips translating only the offending extension (e.g. `RO:0002491(GO:0098768)`) whereas the old code would skip translating the entire ~~annotation~~ _model for the gene_. The error is still being reported out.

### DIFF
--- a/ontobio/rdfgen/gocamgen/gocamgen.py
+++ b/ontobio/rdfgen/gocamgen/gocamgen.py
@@ -601,7 +601,8 @@ class AssocGoCamModel(GoCamModel):
                         if ext_relation == "":
                             # AttributeError: 'NoneType' object has no attribute 'replace'
                             error_msg = "Relation '{}' not found in relations lookup. Skipping annotation translation.".format(str(rel.relation))
-                            raise errors.GocamgenException(error_msg)
+                            self.errors.append(errors.GocamgenException(error_msg))
+                            continue
                         ext_target = str(rel.term)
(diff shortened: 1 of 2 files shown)


kltm:
@dustine32 Just to clarify, previously, as it was a `raise`, an error would cause it to fail and return non-zero, whereas this would mean that it adds to a complaint queue that gets output? Does it return a non-zero status?

dustine32:
Sorry @kltm, one correction to what I said above. The "before" code used to skip translating the entire _model_ (so, _all_ annotations for a gene) if a single extension relation was not in the relations lookup.

To (hopefully) answer your question, the `raise errors.GocamgenException` in the "before" code would be caught at a higher level here:
https://github.com/biolink/ontobio/blob/f00b3d76a3c909ab720d6c13849771bc9a764f6b/ontobio/rdfgen/gocamgen/gocam_builder.py#L89-L90
Essentially this `raise` is a "bail out of translating model" that should then cause to `gpad2gocams` process to simply moves on to translating the next model in a GPAD. This `raise` by itself shouldn't stop the process with a non-zero status and, ideally, the `gpad2gocams` cmd will always return `0` status. The new error handling in this PR shouldn't return a non-zero status either. Hope this helps!

kltm:
Okay, makes sense to me--thank you for the explanation.

kltm (approved):

dustine32:
@kltm Whoa, awesome!

Issue #576, opened 2021-06-15
Run GO rule 20 (obsolete term repair) before GO rule 28 (term aspect check)


dustine32:
Originating from https://github.com/geneontology/pipeline/issues/242#issuecomment-861114921.

Use case here is converting a GPAD, which lacks aspect column, into GAF. For most annotations the null aspect is correctly filled in from the ontology term property `hasOBONamespace`. But this property is missing for obsolete terms (like GO:0044212), so the resulting GAF line will have a blank aspect column. To fix:

> we want to aim to repair the obsoleted term (via GO rule 20) before attempting to extract the aspect in GO rule 28. This way the correct term (GO:0000976) is in place for aspect extraction.

So far, this reordering of rules just looks to involve moving up a few code lines in [gpadparser.py](https://github.com/biolink/ontobio/blob/2066457c5c800e44616440f47f2c499719c87574/ontobio/io/gpadparser.py#L184-L187). We may want to also do this in `gafparser.py` for consistency even though it's likely a non-issue for GAF inputs since the aspects are already explicitly set?

kltm:
@dustine32 I guess this was a problem for https://github.com/geneontology/pipeline/issues/242 , right?
Just to keep everybody at the same place, we can bump go-site `master` to `2.7.6` for ontobio and then have the test branches catch up there.

sierra-moxon:
updated ZFIN branch and rerunning.

PR #577, opened 2021-06-16
Repair obsolete terms before other GO rule tests; for #576

dustine32:
For #576.


kltm:
@dustine32 If I'm reading this correctly, it's flipping how things work a little and a test? If so, take that as my approval.

kltm (approved):

dustine32:
@kltm Correct, it doesn't look like the other tests are dependent on obsolete term repairs occurring in any particular order so should be ok (fingers crossed).

PR #580, opened 2021-07-19
Adding 'existence overlaps' RO relation

vanaukenk:
This is part of the work to transition from using 'exists during' as a GOREL relation to using appropriate temporal relations from RO.

Note that the current entry for 'exists during' refers to RO:0002491 whose label is actually 'existence starts and ends during'.  We should also clean this up at some point, as 'exists during' as a term label currently only exists in gorel.obo for GOREL:0000032.

@dustine32 @kltm - if we updated the entries for 'exists during' (i.e. change to GOREL id) to point to the GOREL id, would that break anything?  I would also then add the correct RO entry for 'existence starts and ends during'.

### DIFF
--- a/ontobio/rdfgen/relations.py
+++ b/ontobio/rdfgen/relations.py
@@ -68,6 +68,7 @@ __relation_label_lookup = bidict({
     "located in": "http://purl.obolibrary.org/obo/RO_0001025",
     "is active in": "http://purl.obolibrary.org/obo/RO_0002432",
     "exists during": "http://purl.obolibrary.org/obo/RO_0002491",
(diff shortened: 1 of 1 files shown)


dustine32:
@vanaukenk I think correcting those other terms in `relations.py` will be fine and shouldn't really break anything. The only thing to look out for is that the GPAD-to-GO-CAM MOD import code uses this lookup as well. So, after the change, running the import code on `exists_during` extensions will translate it to `GOREL:0000032` rather than `RO:0002491`.

BTW, Noctua UI should display the current labels for these (`GOREL:0000032` -> "exists during", `RO:0002491` -> "existence starts and ends during") and isn't directly affected by this `relations.py` lookup.

Issue #582, opened 2021-07-24
Changes to rdfgen/gocamgen for handling new date fields


dustine32:
To implement https://github.com/geneontology/go-shapes/issues/261.

PR #583, opened 2021-08-10
Add handling for creation, modification, and import dates; for #582

dustine32:


dustine32:
Fixes https://github.com/biolink/ontobio/issues/582

kltm:
@dustine32 I think this probably looks good. I was wondering how this handled time, as well as if these files with these additional properties could cause problems for parsing elsewhere? (Not sure how to think about that last part.)

dustine32:
> problems for parsing elsewhere

Good question @kltm! I'm assuming the parsers of these files (Noctua UIs, Minerva, etc.) will ignore the new fields until they're ready, but... perhaps we can wait to merge this until we get some generated models tested in noctua-dev?

As for how the time fields are handled:
* Annotation/line-level
   * `dc:date` - The "modified" date - value comes from `modification-date` annotation property, if present, otherwise regular annotation date (GPAD column 9) value is used.
   * `dct:created` - The "created" date - from `creation-date` annotation property, if present, otherwise no `dct:created` will be emitted.
   * `dct:dateAccepted` - The "import" date - Always just `now.date().isoformat()` at model generation runtime.
* Assertion/axiom/edge-level (since multiple annotations can be collapsed into single assertion individual)
   * `dc:date` - Latest `dc:date` of all annotations collapsed into the single assertion.
   * `dct:created` - Earliest `dct:created` of all annotations collapsed into the single assertion, if present, otherwise no `dct:created` emitted.
   * `dct:dateAccepted` - These should usually be the same date unless a run goes over midnight. If they are different the latest is emitted for the assertion.
*  Model-level
   * `dc:date` - Latest `dc:date` of all annotations in the model.
   * `dct:created` - Earliest `dct:created` of all annotations in the model, if present, otherwise the earliest `dc:date` is emitted.
   * `dct:dateAccepted` Latest `dct:dateAccepted` of all annotations in the model. Again, these should usually all be the same date.

The new `test_model_dates()` in `tests/test_gocamgen.py` might help to illustrate a bit.

kltm:
@dustine32 Great--thank you for the explanation. As long as it is on your mind and we're aware what to watch out for, I'm okay leaving the merge and when to your considerations.

dustine32:
We've [tested models](https://github.com/geneontology/noctua-models/pull/203) in `noctua-dev` generated with these date fields added since 2021-08-10, so merging these changes now.

PR #584, opened 2021-08-12
Update relations.py

vanaukenk:
@dustine32 @kltm @pgaudet

I'm proposing more edits to the relations.py file in anticipation of corrections curators will make to their annotation extensions that currently use the 'exists during' relation.
I've added a line for the RO relation 'existence starts and ends during' and updated the 'exists during' ID to point to the gorel id.
Please let me know if you foresee any issues; I just wanted to add this so we don't start flagging valid AE relations as invalid like we did a few weeks ago for the PomBase annotations.

Thanks.


kltm:
@vanaukenk Is there an issue that we can mark this PR against?

vanaukenk:
@kltm - yes, here it is:

https://github.com/geneontology/go-annotation/issues/2677

kltm:
Personally, looking at the issue, I'm a little unsure of exactly how this fits into the overall picture. That said, these changes seems modest to my eye and would be limited to rdfgen, which is incidental to most of the main data flow at this point. 
I'm comfortable in merging, but would like confirmation from @dustine32 .

dustine32:
@kltm @vanaukenk This lookup is used by the extension parsing (i.e. extension relation labels to CURIEs) in both the GO rules run in the GO pipeline as well as for translating GPAD to GO-CAM. The consequences of this change:

- GO pipeline: should just be that more extensions are accepted (fewer errors causing lines to be filtered out).
- GPAD->GO-CAM: `exists_during` will now translate to the `GOREL:0000032` term rather than `RO:0002491` as before. Relation `existence_starts_and_ends_during` will now be translated and emitted in GO-CAM.

If these consequences are fine with you guys, I think this PR is good to merge!

vanaukenk:
Thanks @dustine32 

These consequences look okay to me, but note that we're trying to get curators to change at the source any 'exists during' extension relations to one of the RO relations, so hopefully for the GPAD->GO-CAM pipeline, we won't have any 'exists during' extensions coming in.  If we do, we should flag them as an error to be checked.  

And actually, in Noctua, 'exists during' is not even a relation option (at least I can't find it in the pop-up window in the graph editor).  Curators would _have_ to choose one of the RO relations in Noctua.

kltm:
Okay, just to be clear, both @dustine32 and @vanaukenk are okay with the changes and consequences. If so (plus or thumbs_up), I'll merge, a release can go out, and we can update our requirements.

PR #585, opened 2021-09-08
Fix bug in obograph_utils by replacing the usage of deprecated networkx methods

deepakunni3:
This PR,
- replaces the usage of a deprecated networkx method `graph.node` with `graph.nodes` in `obograph_util.py`.
- removes the pinning of networkx dependency

@matentzn Thanks for raising this issue again. Hopefully the next release should get rid of the fixed dependency


matentzn:
Awesome!thanks @deepakunni3

kshefchek (approved):

PR #593, opened 2021-10-07
automatically fix obsolete GO terms in the with/from field

sierra-moxon:
Thank you so much for taking the time to look through this!  😁  - happy to refactor.  
first half of fixing: https://github.com/geneontology/go-site/issues/2110

- originally I wanted to call _validate_ontology_class_id directly on with/from fields, unwinding and winding the ConjunctiveSet objects to do that got pretty messy.  
- I also tried/thought I should just do the repair of obsolete terms in the existing str_to_conjugateset method, but it also got a bit messy to call/reuse the logic in _validate_ontology_class_id from str_to_conjugateset. 

I had to fix up a POMBASE GAF parsing test - since the with/from gets passed through the _validate_ontology_class_id method, it warns/reports when a GO id isn't found in the ontology associated with the parser.  I'm not sure why the reported ids wouldn't be in the go-truncated-pombase.json?  

When I run 'make test' locally, I see failing tests that don't show up in the github actions/PR testing.  Should I worry about local test failures like this one? 

```
test_ecomap.py::test_ecomap FAILED                                       [100%]
tests/test_ecomap.py:3 (test_ecomap)
ECO:0007669 != ECO:0000501

Expected :ECO:0000501
Actual   :ECO:0007669
<Click to see difference>

def test_ecomap():
        """
        test mappings between GAF codes and ECO
        """
        m = EcoMap()
        assert m.coderef_to_ecoclass('IEA', 'GO_REF:0000002') == 'ECO:0000256'
>       assert m.coderef_to_ecoclass('IEA') == 'ECO:0000501'
E       AssertionError: assert 'ECO:0007669' == 'ECO:0000501'
```

### DIFF
--- a/ontobio/io/hpoaparser.py
+++ b/ontobio/io/hpoaparser.py
@@ -174,9 +174,8 @@ class HpoaParser(AssocParser):
                 relation = None
 
         # With/From
-        # withfroms = self.validate_pipe_separated_ids(withfrom, split_line, empty_allowed=True, extra_delims=",")
+        withfroms = self.validate_pipe_separated_ids(withfrom, split_line, empty_allowed=True, extra_delims=",")
 
-        withfroms = self._unroll_withfrom_and_replair_obsoletes(split_line, 'gpad')
         if withfroms is None:
             # Reporting occurs in above function call
             return assocparser.ParseResult(line, [], True)


kltm:
@sierra-moxon I'm wondering if it's chance that this is an ECO/IEA error that you're getting, considering recent changes.
I think we need to track down the test issue: either removing it or fixing it. Otherwise, everything seems good to me.
@dustine32 may have more insightful thoughts.

dustine32:
Correct @kltm, that test failure is due to IEA _for reals_ not being `ECO:0000501` anymore (it's now `ECO:0007669`). Not sure what to do with this test on the `EcoMap` functionality, since @cmungall advises against hard-coding ECO codes. Maybe either remove these tests or make a dummy mapping (that won't change) to pass into `ecomap._mappings`?

@sierra-moxon Usually, I'll just run `make travis_test`, which is a smaller set of tests and what travis runs in GH. I believe this `test_ecomap.py` is skipped in `make travis_test`.

Issue #602, opened 2021-11-16
Obsolete GO term with no replacement breaks with/from validation


dustine32:
@sierra-moxon I noticed this error in `_unroll_withfrom_and_replair_obsoletes` when processing a line containing `GO:0004871` in the with/from column:
```
MGI:MGI:2448177         RO:0001025      GO:0005887      MGI:MGI:5467169|PMID:22859307   ECO:0000305     GO:0004871              2013-10-29      MGI             creation-date=2013-10-29|modification-date=2013-10-29|contributor-id=https://orcid.org/0000-0001-7476-6306
```
The error:
```
Traceback (most recent call last):
  File "/Users/ebertdu/go/ontobio/env/bin/ontobio-parse-assocs.py", line 6, in <module>
    exec(compile(open(__file__).read(), __file__, 'exec'))
  File "/Users/ebertdu/go/ontobio/bin/ontobio-parse-assocs.py", line 244, in <module>
    main()
  File "/Users/ebertdu/go/ontobio/bin/ontobio-parse-assocs.py", line 175, in main
    func(ont, args.file, outfh, p, args)
  File "/Users/ebertdu/go/ontobio/bin/ontobio-parse-assocs.py", line 201, in validate_assocs
    for assoc in associations:
  File "/Users/ebertdu/go/ontobio/ontobio/io/assocparser.py", line 521, in association_generator
    parsed_result = self.parse_line(line)
  File "/Users/ebertdu/go/ontobio/ontobio/io/gpadparser.py", line 161, in parse_line
    assoc.evidence.with_support_from = self._unroll_withfrom_and_replair_obsoletes(split_line, 'gpad')
  File "/Users/ebertdu/go/ontobio/ontobio/io/assocparser.py", line 694, in _unroll_withfrom_and_replair_obsoletes
    return association.ConjunctiveSet.str_to_conjunctions(regrouped_fixed_elements)
  File "/Users/ebertdu/go/ontobio/ontobio/model/association.py", line 330, in str_to_conjunctions
    for conj in filter(None, entity.split("|")):
AttributeError: 'NoneType' object has no attribute 'split'
```
Looks like `regrouped_fixed_elements` is `None` when passed into `str_to_conjunctions`:
https://github.com/biolink/ontobio/blob/7ba4e23e55beecf420eb03bdd55e1f0819f5c847/ontobio/io/assocparser.py#L694
[...]

PR #603, opened 2021-11-16
Return empty list for with/from if no term

dustine32:
My fix for the "obsolete GO term having no replaced_by" issue #602.

I need to add a simple test for this to the PR. Thinking we can just piggy-back on: https://github.com/biolink/ontobio/blob/7ba4e23e55beecf420eb03bdd55e1f0819f5c847/tests/test_gpad_parser.py#L13


kltm:
Tests added and passing. Merging.

PR #611, opened 2021-12-20
add pypi publish github action?

sierra-moxon:
To make this work we'd need to talk about adding a pypi API token to this repo (maintainers of ontobio on pypi can do that).


kltm:
@sierra-moxon Where should that conversation happen?

PR #614, opened 2022-02-18
Update qc.py

kltm:
Update to address relaxing GORULE:0000016 for https://github.com/geneontology/go-site/issues/1794


kltm:
Ah, we'll need to make a test correction here as well.

Also, noting that we shouldn't proceed until cleared again: https://github.com/geneontology/go-site/issues/1794#issuecomment-1043707717

kltm:
@dustine32 Great--thank you!

Issue #613, opened 2022-02-02
GO-CAM translation should convert 'transports or maintains localization of' and other relations


dustine32:
Originating from https://github.com/geneontology/go-ontology/issues/22721#issuecomment-1022371319. Specifically:
> MOD imports - add new rule to convert 'transports or maintains localization of' , 'imports', 'exports' to 'has primary input' or 'has primary output' during import

This shakes down to, when emitting annotation extensions, converting these relations:
- transports or maintains localization of (RO:0002313) -> has primary input (RO:0004009)
- imports (RO:0002340) -> has primary input (RO:0004009)
- exports (RO:0002345) -> has primary input (RO:0004009)

@vanaukenk @ukemi Are we going to be converting anything to "has primary output" (RO:0004008)? Feel free to edit above if needed.

dustine32:
Note to self: implementing this conversion/replacement just requires adding lines for "old_relation_label" -> "new_relation_CURIE" here:
https://github.com/biolink/ontobio/blob/34924740c19fc92403ee1dcfa3e34c6725a276ec/ontobio/rdfgen/gocamgen/gocamgen.py#L68-L76
For example:
```
"transports_or_maintains_localization_of": "RO:0004009",
```
We also need to add relations `exports`, `has_primary_input`, and `has_primary_output` missing in `relations.py` [lookup](https://github.com/biolink/ontobio/blob/34924740c19fc92403ee1dcfa3e34c6725a276ec/ontobio/rdfgen/relations.py#L27):
```
"exports": "http://purl.obolibrary.org/obo/RO_0002345",
"has_primary_input": "http://purl.obolibrary.org/obo/RO_0004009",
"has_primary_output": "http://purl.obolibrary.org/obo/RO_0004008",
```

dustine32:
As @vanaukenk referenced in a 2022-03-01 email, the PR https://github.com/geneontology/go-ontology/issues/22962 indicates that all three relations `transports or maintains localization of`, `imports`, and `exports` can be converted to `has primary_input` (RO:0004009). This is reflected in commit https://github.com/biolink/ontobio/commit/6c292715a03aa7fbc0e5d4dfd0d6343b8eeebd33.

dustine32:
This has been implemented and working in the MOD import to GO-CAM translations since late March.

PR #617, opened 2022-03-07
Convert relations to has_primary_input for #613

dustine32:
For #613.

### DIFF
--- a/ontobio/rdfgen/gocamgen/gocamgen.py
+++ b/ontobio/rdfgen/gocamgen/gocamgen.py
@@ -72,7 +72,10 @@ INPUT_RELATIONS = {
     "has input": "RO:0002233",
     "has_input": "RO:0002233",
     "occurs_in": "BFO:0000066",
-    "has_agent": "RO:0002218"
+    "has_agent": "RO:0002218",
+    "transports_or_maintains_localization_of": "RO:0004009",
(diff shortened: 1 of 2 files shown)


kltm:
Why are
`     "transports_or_maintains_localization_of": "RO:0004009",`
and
`     "has primary input": "http://purl.obolibrary.org/obo/RO_0004009",`
so different?

kltm:
More discussion on ticket. Will merge.

dustine32:
@kltm Noting that the first line `"transports_or_maintains_localization_of": "RO:0004009",` is in a dictionary lookup used to transform relations to different relations, e.g. `transports_or_maintains_localization_of` -> `has_primary_input`, in a `gocamgen` annotation-to-model translation context. Whereas the second line `"has primary input": "http://purl.obolibrary.org/obo/RO_0004009",` is the dictionary lookup in `ontobio` used for storing the actual relation label -> CURIE data (essentially, a subset copy of RO). Adding these relations to the latter lookup green-lights their acceptance by the `ontobio` parsers/writers that use `relations.py`.

dustine32:
Linking to explanation about why both `imports` and `exports` are being converted to `has_primary_input`: https://github.com/biolink/ontobio/issues/613#issuecomment-1061204878

Issue #620, opened 2022-05-04
Converting GPAD to GAF can result in blank evidence codes if no ECO-to-GAF mapping exists


dustine32:
Coming from https://github.com/geneontology/go-site/issues/1847 and potentially related to https://github.com/geneontology/pipeline/issues/283.

The `GpadParser` currently allows creation of `GoAssociation` objects with evidence ECO codes that do not [directly map](https://github.com/evidenceontology/evidenceontology/blob/master/gaf-eco-mapping.txt) to a GAF code (e.g. IDA, ISO) so that, when writing these out to GAF, it will result in a blank evidence code column.

We should throw an error on any GPAD line that uses one of these unmapped ECO codes (e.g. `ECO:0006003`). A flag can be created to bypass this check but the default will be to fail these lines.

Tagging @kltm

kltm:
For those reading along: this would generate an error in the report as well, and would not be silent (which is partially the issue now for some cases).

PR #621, opened 2022-05-04
Require GPAD lines to use GAF-compatible ECO codes

dustine32:
For #620.

Added a `--allow_unmapped_eco` flag to `ontobio-parse-assocs.py` to suppress this check for those who don't intend to convert GPAD to GAF. Default is to filter out and report.

I also had to update some test data to use currently valid ECO codes:
```
sed 's/ECO:0000201/ECO:0000266/g' tests/resources/truncated-pombase.gpad
```


kltm (approved):
As discussed--fixes problematic differences between GAF and GPAD output in poorly specified area by limiting it and adding flag to go back to previous behavior. Tests too!

kltm:
@dustine32 Feel free to merge when you feel the time is right.

PR #622, opened 2022-05-06
Lock down evaluating allow_unmapped_eco defaults

dustine32:
For #620.

Quick fix to handle GO pipeline run context as explained here https://github.com/geneontology/go-site/issues/1847#issuecomment-1120020146


kltm:
@dustine32 Rather thoroughly walked me through--looks good.

Issue #623, opened 2022-05-11
Use gaf-eco-mapping-derived.txt in EcoMap in GpadParser and GafWriter


dustine32:
Continuing on the work done in #620:

- [x] In `GpadParser`, ECO codes that are indirectly mapped to a GAF code via an ECO ancestor should be allowed.
- [x] In `GafWriter`, these ECO codes should be converted to their indirectly-mapped GAF codes.

To do this we will use the [gaf-eco-mapping-derived.txt](https://github.com/evidenceontology/evidenceontology/blob/master/gaf-eco-mapping-derived.txt) file in `EcoMap` as needed:
* The `GpadParser` will default to checking for ECO->GAF code mappings in the derived file.
* The `GafWriter` will first check the non-derived [gaf-eco-mapping.txt](https://github.com/evidenceontology/evidenceontology/blob/master/gaf-eco-mapping.txt) for mapping. If none present, it will fall back to the derived file.

kltm:
@dustine32 Great!
Merged https://github.com/biolink/ontobio/pull/625

dustine32:
Tested in GO pipeline `snapshot` and `release` and confirmed this works as intended.

PR #625, opened 2022-05-11
Use gaf-eco-mapping-derived.txt in EcoMap

dustine32:
For #623.


kltm (changes requested):
Adds additional upstream needed (derived); fallback using it; tests.

kltm (approved):

PR #626, opened 2022-07-13
Add additional SO terms to the hard-coded map

kltm:
Not sure why this is hard-coded and not somehow pulling from a dynamic SO, but adding to get moving quickly on https://github.com/geneontology/go-site/issues/1876


kltm:
@dustine32 a little problem here: apparently "gene segment" (note that there is no underscore) is already in use in production, but we have "gene_segment" here and having both causes an error...

PR #627, opened 2022-08-17
remove object from map to prevent accidental removal from fq parameter 

sierra-moxon:
will fix https://github.com/geneontology/helpdesk/issues/285 when go-fastapi becomes production (if we upgraded go-api with this library change it would be fixed as well; best to keep the issue open(?) until go-fastapi goes to prod).


dustine32 (approved):
Straight-forward change.

kltm (approved):
Removes impossible fields

PR #629, opened 2023-02-21
Change IEA expiration date for geneontology/pipeline#317

dustine32:
Implemented this change with @mugitty. We will merge and release to get into GO pipeline.

### DIFF
--- a/.github/workflows/make-tests.yaml
+++ b/.github/workflows/make-tests.yaml
@@ -10,7 +10,7 @@ jobs:
       - uses: actions/setup-python@v2
         name: setup python
         with:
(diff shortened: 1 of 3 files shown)


kltm (approved):
Looks good to me, except for the test failure.

I was wondering what the reason was for the increased specificity of the python version was?

dustine32:
Yeah @kltm the test failure was weird. Turns out to be due to `ubuntu-latest` pointing to a newer ubuntu version that does not have a 3.6 python ready to go. Specifying the ubuntu that we've used before (20.04) seems to have fixed this!

PR #634, opened 2023-05-12
Check if annotated term is in the complex closure

mugitty:
For https://github.com/geneontology/go-site/issues/895 - Check if annotated term is in the complex closure


dustine32 (approved):
Now checks for descendants. Hooray!

PR #635, opened 2023-05-16
For geneontology/go-site#895 fix list bug

mugitty:
For geneontology/go-site#895


dustine32 (approved):
Worked on this with @mugitty so it is good. Code now expects a list of Curie's.

PR #637, opened 2023-05-18
Skip check if ontology is None

mugitty:
For geneontology/go-site#895 - Skip check if ontology is None


dustine32 (approved):
Yes! Allows for cases where ontology not provided in cmd.

PR #638, opened 2023-08-05
For geneontology/go-site#2030

mugitty:
For geneontology/go-site#2030 - Added check for unique taxons


dustine32 (approved):
This makes sense. If an interacting_taxon is specified and everything else is good but interacting_taxon is the same as the object taxon then the QC check fails.

PR #639, opened 2023-08-09
add optional makefile target for generating pyproject.toml and a poetry environment from the existing requirements.txt file

sierra-moxon:
I'd like to do this so that I can manage my python development env for ontobio with poetry.  It is non-breaking to our current .venv process with requirements.txt, and is totally optional for devs.  

pysolr was in the requirements.txt file twice, and dataclasses is a python package that had to be added when using python < 3.7 

(this PR is in prep for editing ontobio GPAD output to match specs as part of this project: https://github.com/geneontology/project-management/issues/77)


kltm:
Reporting as non-breaking optional for development. Why not?

kltm (approved):

kltm:
I'm not sure about the release process, but you can merge when you're ready (unless you want @dustine32 's eyes as well).

sierra-moxon:
> Reporting as non-breaking optional for development. Why not?

why is this a non-breaking change? 

This still assumes requirements.txt is the source of truth for python dependencies and as such, the other kinds of python env managers that use requirements.txt still can, and our package deployment process with setup.py and setup.cfg continue to work as they did before, dependent on requirements.txt

or did you mean, why _not_ make this a breaking change?

dustine32:
@sierra-moxon @kltm I believe the GO pipeline is still running ontobio with (at highest) python3.6. If `dataclasses` not being installed breaks for anything <3.7, we'll need to add `dataclasses` to the go-site/pipeline [reqs.txt](https://github.com/geneontology/go-site/blob/master/pipeline/requirements.txt) file.

Does this sound right?

kltm:
@sierra-moxon Apologies, I was speaking a little shorthand. The intent was to communicate: I don't think this will break anything (non-breaking change) so why not go ahead and merge (combined with my approval).

sierra-moxon:
@dustine32 - I can swap down to 3.6.15 to develop ontobio.  When I do that I get a few errors about packages that are currently in the requirements.txt file that aren't compatible with python versions less than 3.7:

```
The current project's Python requirement (>=3.6,<4.0) is not compatible with some of the required packages Python requirement:
  - click requires Python >=3.7, so it will not be satisfied for Python >=3.6,<3.7
```

```
The current project's Python requirement (>=3.6,<4.0) is not compatible with some of the required packages Python requirement:
  - chardet requires Python >=3.7, so it will not be satisfied for Python >=3.6,<3.7

Because chardet (5.2.0) requires Python >=3.7
 and no versions of chardet match >5.2.0,<6.0.0, chardet is forbidden.
So, because ontobio depends on chardet (^5.2.0), version solving failed.
```

I didn't see any specific dependencies on `chardet` in ontobio, but I do see usages of click....

Do you know where in the pipeline the 3.6 requirement is?  Maybe its ok with 3.7?

dustine32 (approved):

PR #640, opened 2023-08-23
For #844

mugitty:
For https://github.com/geneontology/go-site/issues/844 - Updated to return error if repair obsolete flag is false


dustine32 (approved):
Simple change from warning to error.

PR #641, opened 2023-08-24
Update pypi release instructions

sierra-moxon:


dustine32:
Merging to prevent `__init__.py` version conflict.

PR #642, opened 2023-08-25
For #2038 Changed to error from warning

mugitty:
Updated for GORULE02 change


dustine32 (approved):
Simple change of warn to filter.

PR #643, opened 2023-08-26
For #2049 Added check for DB object symbol

mugitty:
GORULE:0000001 - GAF line should have DB Object symbol in column 3


dustine32 (approved):
Simple enough check to make and I'm kind of surprised it hasn't been done yet.

PR #646, opened 2023-09-14
For #2002 updated due to change in requirements

mugitty:


dustine32:
Linking this PR to https://github.com/geneontology/go-site/issues/2002

ontobio/io/qc.py
```diff
@@ -867,21 +884,16 @@ def test(self, annotation: association.GoAssociation, config: assocparser.AssocP
...
-                        allowed = self.allowed_cc_complex
-                        repair_state = RepairState.FAILED
-            else:
+                    repaired_annotation = copy.deepcopy(annotation)
+                    repaired_annotation.relation = part_of
+                    repaired_annotation.qualifiers = [part_of]
+                    allowed = self.allowed_cc_complex
+                    repair_state = RepairState.REPAIRED
+                    
+            elif term == "GO:0110165" or term in self.make_cellular_anatomical_entity_descendents_if_not_present(config.ontology) or term == "GO:0044423" or term in self.make_virion_component_descendents_if_not_present(config.ontology):
```
dustine32, line 896: With the way the child terms of cellular component are currently set in the ontology:
image
I think this new `elif` conditional is effectively the same as an `else`. Like, any term that is a CC but not a protein-containing complex can be inferred to be either a cellular anatomical entity or a virion component.

It's nitpicky and doesn't really affect anything so I'm still approvin'!

dustine32 (approved):
These changes update the tests. Should be good.

PR #647, opened 2023-10-03
For #2070 check for numeric taxon

mugitty:


dustine32:
Tying this PR to geneontology/go-site#2070

dustine32 (approved):
@mugitty Looks good. Does what it says and there are tests!

PR #648, opened 2023-10-04
Handle GPI label merge for geneontology/go-releases#50

dustine32:
For geneontology/go-releases#50.

This fleshes out the `collections.BioEntities.merge()` function to better handle labels when some are blank. The entity label should only be overwritten if the new label is non-blank.

In our actual use case here:
1. The GAF-derived GPI is parsed into `BioEntities` first.
2. The upstream, canonical MOD GPI is parsed and its own `BioEntities` object.
3. The upstream, canonical MOD GPI `BioEntities` are merged into the existing GAF-derived GPI `BioEntities` ensuring that, if the canonical GPI has a blank symbol, it does not clobber the existing non-blank symbol.

### DIFF
--- a/ontobio/model/collections.py
+++ b/ontobio/model/collections.py
@@ -17,9 +17,21 @@ class BioEntities:
     def merge(self, other):
         """
         Merge another BioEntity set into this one. The `other` set will
-        override any collisions in this BioEntities
+        override any collisions in this BioEntities except for
+        any specific fields (e.g., label) handled below
         """
-        self.entities.update(other.entities)
+        # self.entities.update(other.entities)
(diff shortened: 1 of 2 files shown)


kltm:
@dustine32 By my read, this is a specific `label` clobber override, but could be easily expanded later, correct?

kltm (approved):

kltm:
If https://github.com/biolink/ontobio/pull/648#issuecomment-1747554581, feel free to merge on your schedule.

dustine32:
@kltm Yup, that block is there for any future field-merge logic with label being the only current use.

PR #649, opened 2023-10-07
Reverting FAIL to OKAY since obsolete terms have no namespace

dustine32:
Change corresponds to geneontology/go-site#2002. Slight correction of initial PR https://github.com/biolink/ontobio/pull/646.

This fixes a bug introduced with https://github.com/biolink/ontobio/pull/646 that was failing all obsoleted terms in rule 61 because their ontology node did not have a `namespace` key:
https://github.com/biolink/ontobio/blob/6a9924a1295f896dfdf6df5796a9e1b6d1203df9/ontobio/io/qc.py#L907
An example obsoleted term is GO:1902361. The failed GO rule test example is here: https://github.com/geneontology/go-site/blob/master/metadata/rules/gorule-0000020.md

I added a test for this (within `ontobio`!!) under rule 61 and confirmed that the `go-site` [test suite](https://github.com/geneontology/go-site/blob/master/scripts/rule-example-validation.sh) (really just a wrapper around `ontobio`'s `validate.py rule` command) runs w/o error.

### DIFF
--- a/ontobio/io/qc.py
+++ b/ontobio/io/qc.py
@@ -904,7 +904,8 @@ class GoRule61(RepairRule):
         else:
             # If we reach here, we're in a weird case where a term is not in either
             # of the three main GO branches, or does not have a namespace defined.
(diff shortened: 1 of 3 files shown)


kltm (approved):
Fix and test.

PR #650, opened 2023-10-12
For #1658

mugitty:
Updated qualifier labels


dustine32:
Linking to ticket https://github.com/geneontology/go-site/issues/1658

mugitty:
@dustine32, the WB line fails the qualifier test as expected. 

@pgaudet how should be proceed?

pgaudet:
@dustine32 

What is the correct way format for these these, with or without comma? We could have a REPAIR for these. 
In RO and in GOREL there is a comma, but we can remove it, if groups don't use it.

dustine32:
@pgaudet @mugitty I'd lean on whichever format is in the [GAF 2.2 spec](https://geneontology.org/docs/go-annotation-file-gaf-format-2.2/#qualifier-column-4) as I assume this is what everyone is currently following to produce their GAFs. This spec lists out accepted qualifier values and none of them contain a comma.

Though, I'm still a bit uncertain who has authority over these GAF qualifier values. @kltm @vanaukenk @balhoff These underscored qualifier values (e.g., `acts_upstream_of_or_within_negative_effect`, `is_active_in`) are just shorthand we use at GO to translate from GAF to the proper RO, GOREF, etc. CURIE, which then will be tied to a proper label (e.g., "acts upstream of or within, positive effect", "is active in") maintained at their respective ontologies. Does this sound right?

In other words, there are three distinct Term fields here:
* Term IRI - RO:004032
* Term label - acts upstream of or within, positive effect
* Term GAF qualifier - acts_upstream_of_or_within_negative_effect

And we at GO can just demand (via the GAF spec) that GAF producers use whatever GAF qualifiers (i.e., w/ comma or w/o comma) we like?

kltm:
> Though, I'm still a bit uncertain who has authority over these GAF qualifier values. @kltm @vanaukenk @balhoff These underscored qualifier values (e.g., `acts_upstream_of_or_within_negative_effect`, `is_active_in`) are just shorthand we use at GO to translate from GAF to the proper RO, GOREF, etc. CURIE, which then will be tied to a proper label (e.g., "acts upstream of or within, positive effect", "is active in") maintained at their respective ontologies. Does this sound right?

While the format is the format, we have control over expectations and QC. Once upon a time, the labels were pretty much just that, but became tied to ontology terms behind the scenes as time went on. The GAF still displays this history on its shirt, GPAD/GPI does away with it. Different consumers of the files treat these different (or ignores them), but we are ideally tying everything back to the appropriate ontology term when we process.

> In other words, there are three distinct Term fields here:
> 
>     * Term IRI - RO:004032

<cough>CURIE<cough>

>     * Term label - acts upstream of or within, positive effect

Yes.

>     * Term GAF qualifier - acts_upstream_of_or_within_negative_effect

Unfortunate token proxy for ontology term

> And we at GO can just demand (via the GAF spec) that GAF producers use whatever GAF qualifiers (i.e., w/ comma or w/o comma) we like?

Well, the ones defined in the spec. That's the rule. We map those to the appropriate term.

mugitty:
According to https://geneontology.org/docs/go-annotation-file-gaf-format-2.2/#qualifier-column-4, the commas are not in the qualifiers.  Therefore, update for commas is not required

pgaudet:
We dont want the comma anymore.- please close

PR #651, opened 2023-10-13
For #2079 added check for binding

mugitty:


dustine32 (approved):
@mugitty Straight-forward change so good-to-go!

dustine32:
Linking to issue https://github.com/geneontology/go-site/issues/2079

PR #654, opened 2023-10-20
For #2081 repair goid, with/from and extension before running gorule …

mugitty:
…suite

### DIFF
--- a/ontobio/io/gafparser.py
+++ b/ontobio/io/gafparser.py
@@ -201,55 +201,8 @@ class GafParser(assocparser.AssocParser):
         if self.gaf_version() == "2.1" and (vals[3] == "" or vals[3] == "NOT") and self.config.ontology:
             assoc = self.upgrade_empty_qualifier(assoc)
 
-        ## Run GO Rules, save split values into individual variables
-        # print("Config is {}".format(self.config.__dict__.keys()))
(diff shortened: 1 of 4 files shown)


kltm (approved):
Looks good eyeballing it--I'll leave it to you and @dustine32 to determine the right time to merge and release.

Is there a test that was false that will now be true with this change?

mugitty:
The test that was previously failing is part of the ones in @pgaudet's list.  I will add to gorule 6 test cases.

dustine32:
@mugitty Thanks for fixing this! I can't think of any reason we needed the GO rules to run before the GO term obsolete repair so this should be good.

However, have you tried running Pascale's [tests](https://github.com/geneontology/go-site/blob/fe5452121d10b71264c0dc58dcf07e8813843167/scripts/rule-example-validation.sh#L7) in `go-site` (the `validate.py rule` command) to confirm this fixes the issue?

mugitty:
@dustine32 , I ran it with Pascale's errors.gaf file and the validate.py rule. I want to add a test case where the GO id is fixed and checked by GORULE 6. I will let you know when it is ready

Thanks

mugitty:
Code updated.  Please review

dustine32 (approved):
Went through these code changes with @mugitty. Tests all pass so we are good!

PR #656, opened 2023-11-09
For #1523 check for unique GO ID

mugitty:
Updated to ensure that annotations with evidence code 'IC' have a GO Id entry in the 'with/from' field that is different from the GO Id for the annotation

### DIFF
--- a/ontobio/io/qc.py
+++ b/ontobio/io/qc.py
@@ -352,15 +352,16 @@ class GoRule15(GoRule):
 class GoRule16(GoRule):
 
     def __init__(self):
-        super().__init__("GORULE:0000016", "All IC annotations should include a GO ID in the \"With/From\" column", FailMode.SOFT)
+        super().__init__("GORULE:0000016", "All IC annotations should include a GO ID in the \"With/From\" column that is also different from the entry in the \"GO ID\" column", FailMode.SOFT)
 
     def test(self, annotation: association.GoAssociation, config: assocparser.AssocParserConfig, group=None) -> TestResult:
         evidence = str(annotation.evidence.type)
         withfrom = annotation.evidence.with_support_from
+        goId = annotation.object.id
 
         okay = True
         if evidence == ic_eco:
-            only_go = [t for conjunctions in withfrom for t in conjunctions.elements if t.namespace == "GO"] # Filter terms that aren't GO terms
+            only_go = [t for conjunctions in withfrom for t in conjunctions.elements if (t.namespace == "GO" and goId is not None and t.identity != goId.identity)] # Filter terms that aren't GO terms and different from GO ID
             okay = len(only_go) >= 1
 
         return self._result(okay)
--- a/ontobio/io/qc.py
+++ b/ontobio/io/qc.py
@@ -357,11 +357,12 @@ class GoRule16(GoRule):
     def test(self, annotation: association.GoAssociation, config: assocparser.AssocParserConfig, group=None) -> TestResult:
         evidence = str(annotation.evidence.type)
         withfrom = annotation.evidence.with_support_from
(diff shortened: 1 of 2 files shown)


ontobio/io/qc.py
```diff
@@ -352,15 +352,16 @@ def test(self, annotation: association.GoAssociation, config: assocparser.AssocP
 class GoRule16(GoRule):
 
     def __init__(self):
-        super().__init__("GORULE:0000016", "All IC annotations should include a GO ID in the \"With/From\" column", FailMode.SOFT)
+        super().__init__("GORULE:0000016", "All IC annotations should include a GO ID in the \"With/From\" column that is also different from the entry in the \"GO ID\" column", FailMode.SOFT)
 
     def test(self, annotation: association.GoAssociation, config: assocparser.AssocParserConfig, group=None) -> TestResult:
         evidence = str(annotation.evidence.type)
         withfrom = annotation.evidence.with_support_from
+        goId = annotation.object.id
```
dustine32, line 360: This is picky but can `goId` be called `go_id` instead to stay consistent with python's underscore [convention](https://peps.python.org/pep-0008/#method-names-and-instance-variables)?

ontobio/io/qc.py
```diff
@@ -352,15 +352,16 @@ def test(self, annotation: association.GoAssociation, config: assocparser.AssocP
...
 
         okay = True
         if evidence == ic_eco:
-            only_go = [t for conjunctions in withfrom for t in conjunctions.elements if t.namespace == "GO"] # Filter terms that aren't GO terms
+            only_go = [t for conjunctions in withfrom for t in conjunctions.elements if (t.namespace == "GO" and goId is not None and t.identity != goId.identity)] # Filter terms that aren't GO terms and different from GO ID
```
dustine32, line 364: Is the `goId is not None` criteria necessary here? If `annotation.object.id` is blank this annotation is already in pretty bad shape and would definitely fail/crash somewhere else.

tests/test_qc.py
```diff
@@ -247,9 +247,14 @@ def test_go_rules_15():
     assert test_result.result_type == qc.ResultType.WARNING    
 
 def test_go_rule_16():
+    # GO term same as with/ID
+    assoc = make_annotation(goid="GO:0044419", evidence="IC", withfrom="GO:0044419").associations[0]
```
dustine32, line 251: This test is good and I think another good test would be to include multiple GO terms in the test `withfrom`, one valid (differs from `goid`) and one invalid (same) to ensure it's not dropping `withfrom` entirely. So:
```
assoc = make_annotation(goid="GO:0044419", evidence="IC", withfrom="GO:0044419|GO:0035821").associations[0]
```
This should return `qc.ResultType.PASS` because the with/from still has >=1 valid GO term.

dustine32 (changes requested):
@mugitty Very minor/picky things to check in the code comments.

ontobio/io/qc.py
mugitty: Good point.  Will update

tests/test_qc.py
mugitty: Good point.  Also code should output message even if one of the terms is same as go id for annotation

ontobio/io/qc.py
mugitty: ok

mugitty:
@dustine32 please review

dustine32 (approved):
@mugitty Thanks for the changes and for investigating the (non-existent) primary GO term checks! It looks like your fix for https://github.com/geneontology/go-site/issues/2175 will ensure that `go_id.namespace` always equals `"GO"` in this test.

PR #658, opened 2023-11-10
For #2124 Disable Gorule51

mugitty:


dustine32 (approved):
All good here. As @mugitty and I discussed on slack, removing `GoRule51` from the `GoRules` Enum should prevent it from running by default.

dustine32:
Tying PR to https://github.com/geneontology/go-site/issues/2124.

PR #659, opened 2023-11-14
For #2131

mugitty:
Updated reference for GoRule37


dustine32 (approved):
@mugitty Thanks for getting this quick!

PR #660, opened 2023-11-21
null check with_from items

sierra-moxon:


kltm:
https://github.com/geneontology/pipeline/issues/346

dustine32:
@mugitty is going to push an update to this `qc_null_check` branch to get the failing `test_go_rule_16` test to pass.

mugitty:
After discussing with @dustine32, we decided to make additional changes. I have committed additional changes

ontobio/io/gpadparser.py
mugitty, line 1: Failure of gorule16 is soft.  Need to ensure assoc.evidence.with_support_from is not None before traversing

ontobio/io/qc.py
mugitty, line 1: Gorule16 needs to verify that for annotation with code IC, at least one of the with/from is a go term that is not same as the go term for the annotation

dustine32 (approved):
@mugitty Looks good to me. The travis tests and `validate.py rule` all pass so merge whenever.

tests/test_qc.py
mugitty, line 1: Added test case for Gorule16 where GPAD parser creates annotation with evidence code ECO that does not have a with/from

PR #661, opened 2023-11-28
For #1658

mugitty:
Updated for MF term accession


dustine32 (approved):
@mugitty I think this a straight-forward change to make.

PR #662, opened 2023-12-01
For #346

mugitty:
Return error if with/from is null


dustine32 (approved):
@mugitty This should fix the pipeline error. I've retested with our example MGI line having with/from = "GO:0031061".

PR #664, opened 2024-01-12
remove -aux from golr url

sierra-moxon:


dustine32 (approved):
If this makes sense to @sierra-moxon then it makes sense to me!

PR #667, opened 2024-02-22
For #1553

mugitty:
Updated for deprecated gorule

### DIFF
--- a/ontobio/io/qc.py
+++ b/ontobio/io/qc.py
@@ -687,35 +687,6 @@ class GoRule55(GoRule):
         return self._result(True)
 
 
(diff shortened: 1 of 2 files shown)


dustine32:
Tying this to geneontology/go-site#1553.

dustine32 (approved):
Straight-forward removal of GoRule code is good with me!

@mugitty Just curious. Does this automatically remove GO rule 57 from the `assigned-by-gorule-report.html` or is this controlled in `go-site`?

mugitty:
@dustine32 , go-site does not control it.  Once ontobio stops outputting information about the rule, the reports will not output information about them.

PR #672, opened 2024-03-28
For #2273

mugitty:
Implemented gorule-0000064

### DIFF
--- a/bin/ontobio-parse-assocs.py
+++ b/bin/ontobio-parse-assocs.py
@@ -143,9 +143,11 @@ def main():
         rule_set = assocparser.RuleSet.ALL
     
     goref_metadata = None
+    goref_species = None
     if args.metadata_dir:
         absolute_metadata = os.path.abspath(args.metadata_dir)
-        goref_metadata = metadata.yamldown_lookup(os.path.join(absolute_metadata, "gorefs"))    
+        goref_metadata = metadata.yamldown_lookup(os.path.join(absolute_metadata, "gorefs"))
+        goref_species = metadata.yaml_set(absolute_metadata, "go-reference-species.yaml", "taxon_id")
 
     # set configuration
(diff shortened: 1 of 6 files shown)


kltm (commented):
Quick review looks good to me, but I'd like to defer the final call/merge/release to @dustine32

bin/ontobio-parse-assocs.py
```diff
@@ -143,9 +143,11 @@ def main():
         rule_set = assocparser.RuleSet.ALL
     
     goref_metadata = None
+    goref_species = None
     if args.metadata_dir:
         absolute_metadata = os.path.abspath(args.metadata_dir)
-        goref_metadata = metadata.yamldown_lookup(os.path.join(absolute_metadata, "gorefs"))    
+        goref_metadata = metadata.yamldown_lookup(os.path.join(absolute_metadata, "gorefs"))
+        goref_species = metadata.yaml_set(absolute_metadata, "go-reference-species.yaml", "taxon_id")
```
dustine32, line 150: My nitpicky thing here that doesn't need to be fixed (and I'm very open to arguing whether it's actually wrong) is the naming of the "goref_species" variable. I think its name is derived from `goref_metadata` above, which points to GO_REF ID data. However, the species data is totally different from GO_REFs and I might be misled into thinking the `goref_species` var has something to do with GO_REF IDs.

Could we call it something like `reference_species_metadata` or `ref_species_metadata`? @mugitty What do you think? Ignore this if you like!

dustine32 (approved):
Looks good @mugitty. Will filter _out_ species in our reference species list.

bin/ontobio-parse-assocs.py
mugitty: @dustine32, I have updated the variable name

dustine32: Thanks so much @mugitty!

PR #673, opened 2024-03-29
For #2273 Fix path

mugitty:


dustine32:
Thanks @mugitty! I'm actually running a local test of `go-site-2273-path-fix` with `validate.py rule` and `produce` right now and then will review if it succeeds. Quality control!

dustine32 (approved):
@mugitty Thanks for the quick fix! I confirmed that running `rule` and `produce` locally do not err out.

dustine32:
@mugitty Merge away!

mugitty:
Thanks for reviewing and testing

PR #674, opened 2024-05-09
Go site 676 gorule 0000022 check for retracted publications

mugitty:


kltm (changes requested):
@mugitty These changes generally make sense to me.
Looking at this, one thing we had discussed was making the use of retracted publications completely optional--i.e. adding a new option to the path and only executing if available. It looks like this just depends on the go-site metadata directory?

kltm:
@mugitty It looks like the test error:
```
==================================== ERRORS ====================================
_______________ ERROR collecting tests/test_validation_rules.py ________________
tests/test_validation_rules.py:1: in <module>
    from ontobio.validation import rules
ontobio/validation/rules.py:11: in <module>
    from ontobio.validation import metadata
ontobio/validation/metadata.py:188: in <module>
    def retracted_pub_set(metadata)->set[str]:
E   TypeError: 'type' object is not subscriptable
=============================== warnings summary ===============================
```
is related to the handling of go-site metadata as mentioned above.

mugitty:
@kltm , I did not realized that we wanted a flag to determine if the test should be run.  So, it would only be in ontobio-parse-assocs.py as a parameter?

For the error, is it expecting set[] instead of set[str] due to version of python?

kltm:
@mugitty For me, I just want to make sure that the file is optional, and ideally be able to specify it in different contexts. The exact mechanism isn't too too important.

dustine32:
Linking to issue geneontology/go-site#676

kltm (commented):
Examining code, it looks like:

- if file given, rule run
- if file found at assumed location, rule run
- otherwise, rule skipped

I believe this is in line with what we discussed; if so, please merge at your convenience

PR #677, opened 2024-06-24
Go site 2210 gorule 0000027 must check dbs are in the db xref file

mugitty:


dustine32:

ontobio/io/gpadparser.py
```diff
@@ -235,6 +239,8 @@ def parse_line(self, line):
         references = self.validate_curie_ids(assoc.evidence.has_supporting_reference, split_line)
```
dustine32, line 239: @mugitty Is there any overlap of logic between `self.validate_curie_ids` here and `self._validate_curie_using_db_xrefs` just below? Could `self._validate_curie_using_db_xrefs` be incorporated into `self.validate_curie_ids`?

dustine32:
@mugitty Do we already have a test for gorule 0000027 somewhere?

ontobio/io/gpadparser.py
mugitty: @dustine32 , validate_curie_ids calls _validate_id . This checks for things like "DB:id" or for annotations, prefix has to be in GO id space.  However, it does not validate against the syntax pattern specified in the db-xrefs file. This is meant as a catch-all for any identifier in the GAF line that has a database field and id_syntax

dustine32: @mugitty Would there be any issue if you incorporated this db-xrefs syntax pattern checking inside `_validate_id`? It [looks](https://github.com/biolink/ontobio/blob/3576a13929f01cd243b6401e2825405b5401e0b5/ontobio/io/assocparser.py#L830) like `_validate_id` assumes every `id` is a CURIE, which by definition should always have colon-separated database field and id_syntax (else an error will be reported) so both . The only complication I could think of is if a `metadata/db-xrefs` is not supplied when `_validate_id` is called but you could just make this optional in `_validate_id` (i.e., `if self.config.db_type_name_regex_id_syntax is not None`).

I guess my main concern is separating validation logic throughout the code that really should be in the same place.

mugitty: @dustine32, as you suggested, I could incorporate in _validate_id and add a check for self.config.db_type_name_regex_id_syntax is not None.  

Let me update

mugitty:
@dustine32 , I pushed updates to combine id syntax change into _validate_id method

dustine32:
@mugitty Awesome, thank you so much! Do we have a test for wrong IDs that should return that "does not match any id_syntax patterns" warning message? It would be good to confirm your new code can be triggered by this.

mugitty:
I ran it through https://github.com/geneontology/go-site/blob/master/docs/gorules_test_errors.gaf . The output error.json file contains things such as 
...
            {
                "level": "WARNING",
                "line": "MGI:1100518\tSmad7\tbla\tinvolved_in\tGO:0017015\tMGI:MGI:3836072|PMID:18952608\tIC\tGO:0060389\tP\tGORULE_TEST:0000020-3\tSMAD\tprotein_coding_gene\ttaxon:10090\t20090211\tGO_Central\t\n",
                "type": "Invalid identifier",
                "message": "GORULE:0000027: 3836072 does not match any id_syntax patterns for MGI in dbxrefs",
                "obj": "MGI:MGI:3836072",
                "taxon": "NCBITaxon:10090",
                "rule": 27
            },
            {
                "level": "WARNING",
                "line": "UniPotKB\tQ9HC96\tCAPN10\tinvolved_in\tGO:0006921\tPMID:23072806\tIDA\t\tP\tGORULE_TEST:0000027-1 Calpain-10\tCAPN10,KIAA1845\tprotein\ttaxon:9606\t20140213\tGO_Central\t\n",
                "type": "Invalid identifier",
                "message": "GORULE:0000027: UniPotKB not found in list of database names in dbxrefs",
                "obj": "UniPotKB:Q9HC96",
                "taxon": "NCBITaxon:9606",
                "rule": 27
            },
            {
                "level": "WARNING",
                "line": "UniProtKB\tQ9HC96\tCAPN10\tinvolved_in\tGO:0006921\tPMID:PMID:14561399\tIDA\t\tP\tGORULE_TEST:0000027-3 Calpain-10\tCAPN10,KIAA1845\tprotein\ttaxon:9606\t20140213\tGO_Central\t\n",
                "type": "Invalid identifier",
                "message": "GORULE:0000027: PMID:14561399 does not match any id_syntax patterns for PMID in dbxrefs",
                "obj": "PMID:PMID:14561399",
                "taxon": "NCBITaxon:9606",
                "rule": 27
            },
            ...
            
 This update outputs warnings.  Groups can update the id_syntax and or ids based on these errors

dustine32:
@mugitty Thanks again! It's great that you confirmed the `go-site` checks will run correctly with this change but usually we also test this type of functionality within the ontobio/tests. As a comparable example, I'm looking at the test for [GoRule43](https://github.com/biolink/ontobio/blob/3576a13929f01cd243b6401e2825405b5401e0b5/tests/test_qc.py#L571) and it recreates a small part of the gorefs metadata to test the GO_REF validation functionality. Could we do the same thing in `test_qc.py` (i.e., add a `test_gorule27` function) with a small sample of the `db-xrefs` metadata?

mugitty:
@dustine32, since these are validation tests,  added tests to both test_gafparser.py and test_gpad_parser.py

dustine32 (approved):
@mugitty Thanks so much for your patience and thanks for adding the tests! This all looks good now.

PR #678, opened 2024-07-15
For #2210

mugitty:


dustine32:

dustine32 (approved):
@mugitty Confirming this change is specifically to enable the retracted publication check? I can see how this would fix it by passing the required variables from `produce` to `produce_gaf`.

PR #679, opened 2024-07-19
For #2210

mugitty:


dustine32 (approved):
@mugitty Moving from `regex.match` to `regex.fullmatch`. I assume this is intended to be more strict to catch some invalid values that were previously getting through the `.match` function?

dustine32:

mugitty:
> @mugitty Moving from `regex.match` to `regex.fullmatch`. I assume this is intended to be more strict to catch some invalid values that were previously getting through the `.match` function?
Yes @dustine32 this will make it more restrictive since the pattern should match the entire curie string

PR #681, opened 2024-07-23
swaps subject with gene when subject is PR isoform

sierra-moxon:
fixes https://github.com/geneontology/go-site/issues/2291


kltm (approved):
Small fixes, tests, small cleaning; mainly "fixing" isoform/gp representation being differentially treated in GPAD vs GAF.

PR #683, opened 2024-07-25
force GAF 2.2 output, instead of taking version from input GAFs - fixing the chickens

sierra-moxon:


bin/validate.py
```diff
@@ -787,7 +801,7 @@ def fix_pro_isoforms_in_gaf(gaf_file_to_fix: str, gpi_file: str, ontology_graph,
                                         "id": gpi_entry.get('id')}
 
     gafparser = GafParser(config=assocparser.AssocParserConfig(ontology=ontology_graph))
-    gafwriter = GafWriter(file=open(output_file_path, "w"), source="test", version=gafparser.version)
+    gafwriter = GafWriter(file=open(output_file_path, "w"), version="2.2")
```
sierra-moxon, line 804: @dustine32 @kltm—this is the biggest change in this PR: Instead of taking the src-input GAF version to populate the output GAF, write out version 2.2.  

The rest of the PR is helpful too - e.g. creating temporary files instead of over-writing files, some test fixes of headers (e.g. consistently use GOC instead of a mix of "GOC" and "GO Central"), but this line is what fixed the chickens.

bin/validate.py
```diff
@@ -711,10 +711,7 @@ def produce(ctx, group, metadata_dir, gpad, gpad_gpi_output_version, ttl, target
                                 rule_set=rule_set
                                 )[0]
 
-        click.echo("Producing GPI from GAF files...")
-        gpi = produce_gpi(dataset, absolute_target, valid_gaf, ontology_graph, gpad_gpi_output_version)
-
-        gpi_list = [gpi]
+        gpi_list = []
```
dustine32, line 714: @sierra-moxon Just curious why `gpi_list` is now initialized w/o the GAF-source GPI. The plumbing might have changed to make this step unnecessary, maybe? I know I plugged this in as part of https://github.com/geneontology/pipeline/issues/239 (Noctua GPAD entities not in any GPI were getting `taxon:0`) so hopefully the tests would scream if this broke anything.

kltm (commented):
I'll defer to @dustine32 for this. Feel free to merge (release and update go-site/pipeline/requirements.txt) when it looks right.

bin/validate.py
sierra-moxon: yeah great catch - let me know if this logic is correct:

- initialize empty GPI list
- get the GPI from the dataset metadata if available and add it to the GPI list
- get noctua GPAD (if exists)
- get the PAINT GAF (if exists)
- make GPI using the valid_gaf (which is the GAF file that has passed thru a rule check) and add it to the GPI list
- make the GPAD using valid_gaf, noctua GPAD, PAINT GAF, and the list of GPIs in the GPI list
- make a GAF incorportating noctua, paint annotations.
- run the resulting GAF through the isoform fixer and make a "final" GAF (this is where we swap `PR:` IDs for their gene id equivalent). 
- make the final GPI file **using this final GAF file**

The last step I did so that we weren't creating a GPI file that was missing content based on additions along the way here (e.g. GPI creation after noctua and PAINT incorporation and after the isoform swap).  Is this the right logic?  

I will run through my manual test with ZFIN here and check that no NCBITaxon:0 nor taxon:0 values are found.

dustine32: @sierra-moxon Thanks for writing all of that out! I believe that should work. Especially since I now don't think the GAF-sourced GPI was the crucial data we needed for the original [issue](https://github.com/geneontology/pipeline/issues/239) - we really needed the upstream MOD GPI (e.g., for ZFIN) in `gpi_list` and your changes still have that.

dustine32 (approved):
@sierra-moxon Thanks for explaining!

PR #684, opened 2024-07-29
fixes pipeline issue 383 - gpad 1.2 should have `!gpa-version` in the header, not `!gpad-version` in the header

sierra-moxon:


dustine32 (approved):
For geneontology/pipeline#383.

@sierra-moxon Straightforward change. This should fix the header field issue for early GPAD versions. Thanks!

PR #686, opened 2024-08-14
Pypi publish

sierra-moxon:
@dustine32 - I did see your comment in my other PR here that tries to isolate tests into the poetry environment which I still need to work on replicating. It might take me a couple of tries here in case of a config issue that I forgot... if you're willing to bear with me here to do some fussing.... 🫠 😄


kltm (approved):
The changes here--adding the publish step--seem sensible to me.
That said, you know more about this than I, so I'm happy for you to proceed as you see fit.

kltm:
(Naturally pending passing tests.)

PR #687, opened 2024-08-16
revert autoversion, set up setup.py for a manual push to pypi

sierra-moxon:


dustine32 (approved):
Simple delete of the neat GH action workflow file. Too bad.

PR #692, opened 2024-09-06
For #2360 - Updated to handle DBs without id syntax pattern

mugitty:


dustine32:
Linking to https://github.com/geneontology/go-site/issues/2360

dustine32 (approved):
@mugitty Ok, sorry this took so long for me to debug (thanks PyCharm!) and understand but I think this is good.

Basically, this now allows any prefix that's in dbxrefs.yaml to be a valid ID if it doesn't have any special ID syntax patterns. Whereas before, `identity_matches_pattern` would be `False` and attempting to iterate an empty dictionary of `type_name_regex_patterns` would be unable to set `identity_matches_pattern` to `True`, causing the ID to be considered invalid.

mugitty:
@dustine32, previously any db-xref's without id_syntax specifier were being flagged as being invalid. This update will permit db-xrefs without id_syntax specifier

PR #698, opened 2024-11-12
Issue pipeline 387 - add GAF header information to aggregated GPAD produced by megamake step (aka. ontobio/bin/validate.py) 

sierra-moxon:
see: https://github.com/geneontology/pipeline/issues/387#issuecomment-2471589993


dustine32 (approved):
Tests pass and I think I like how the write-out is now separated from the `process_[gaf|gpad]...` functions now.

PR #699, opened 2024-12-09
refactor method to use .get() and check for property availablity befo…

### LINKED ISSUE #697: problems parsing output from latest obographs library
In GO, we are trying to update to the latest obographs in ROBOT. We have been using owltools, which uses an older obographs release. We noticed this change, which currently upsets the ontobio JSON reader:

owltools:

```json
{
      "definedClassId" : "http://purl.obolibrary.org/obo/GO_0101013",
      "genusIds" : [ "http://purl.obolibrary.org/obo/GO_0005248", "http://purl.obolibrary.org/obo/GO_0008381" ],
      "restrictions" : [ ]
}
```

robot:

```json
{
      "definedClassId" : "http://purl.obolibrary.org/obo/GO_0101013",
      "genusIds" : [ "http://purl.obolibrary.org/obo/GO_0005248", "http://purl.obolibrary.org/obo/GO_0008381" ]
}
```

owltools outputs an empty `restrictions` array, but the newer obographs leaves it out. As pointed out by @cmungall in [my issue at obographs](https://github.com/geneontology/obographs/issues/110), all properties in a JSON schema are optional unless marked required. So ontobio should be able to handle a missing `restrictions` and treat it as an empty list. I think there are other properties that will trigger this, so the same fix should be applied to all list handling.

cc @kltm @sierra-moxon

sierra-moxon:
…re processing fixes #697 

refactor the convert_json_object method of obograph_utils to handle potentially null inputs to non-required properties.


dustine32 (approved):
@sierra-moxon This looks good to me. No longer chokes on `a['restrictions']` if `'restrictions'` key not present.

Thanks!

PR #702, opened 2024-12-13
For geneontology/pipeline#408

mugitty:


dustine32 (approved):
This change correctly matches the right part of an MGI ID ex: "MGI:12345" vs "12345" against the db-xref pattern entry. Code change looks good and tests work. Thanks @mugitty!

PR #703, opened 2024-12-13
For geneontology/go-site#2392

mugitty:


dustine32 (approved):
Yup, similar to GoRule08. This checks a term for `gocheck_obsoletion_candidate` subset and fails annotation if yes.

